PR #2, opened 2018-05-31
Implement json to stix

Logicon211:


README.md
```diff
@@ -16,6 +16,14 @@ Returns:
 
 `SELECT * FROM events WHERE destinationip='192.168.122.84' OR sourceip='192.168.122.84' OR destinationip='192.168.122.83' OR sourceip='192.168.122.83'`
 
+Example of converting AQL events to STIX:
+
+`python3 stix_shifter.py "qradar" "qradar_events" '[{"starttime": 1524227777191, "protocolid": 255, "sourceip": "9.21.123.112", "logsourceid":126, "qid": 55500004, "sourceport": 0, "eventcount": 1, "magnitude": 4, "identityip": "0.0.0.0", "destinationip": "9.21.123.112", "destinationport": 0, "category": 10009, "username": null}]'`
```
delliott90: I try to just keep this as python instead of python3 even though that's what I need to use on my mac for things to work. I think if the python 3 is set up properly it's not needed.

Logicon211: Right good point, I'll update that

README.md
```diff
@@ -16,6 +16,14 @@ Returns:
...
+
+Returns:
+
+`[{'x_com_ibm_uds_datasource': {'id': '7c0de425-33bf-46be-9e38-e42319e36d95', 'name': 'events'}, 'id': 'observed-data--600ddcf4-8d21-4667-955a-2d14e157b1f2', 'type': 'observed-data', 'objects': {'0': {'type': 'ipv4-addr', 'value': '9.21.123.112'}, '1': {'type': 'ipv4-addr', 'value': '9.21.123.112'}}, 'x_com_ibm_ariel': {'log_source_id': 126, 'identity_ip': '0.0.0.0', 'protocol_id': 255, 'magnitude': 4, 'qid': 55500004}, 'number_observed': 1, 'created': '2018-04-20T12:36:17.191Z', 'modified': '2018-04-20T12:36:17.191Z', 'first_observed': '2018-04-20T12:36:17.191Z', 'last_observed': '2018-04-20T12:36:17.191Z'}]`
```
delliott90: Is it possible in markdown to keep the proper JSON formatting instead of it wrapping?

Logicon211: Do you mean in the actual return value or just in this readme?

I can definitely update the readme to look nice. The actual return from the library won't be so nicely formatted without a bit of work though (And it probably wouldn't be a good idea since formatting it might mess with whoever is going to consume this)

src/json_to_stix/transformers.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+from . import json_to_stix
+from datetime import datetime, timezone
+import pprint
+import logging
+logging.basicConfig(format='%(asctime)s %(message)s', level=logging.DEBUG)
+
+pp = pprint.PrettyPrinter(indent=2, width=50)
```
delliott90: Do we even use this variable and the above pretty print? I can't see where it gets referenced.

Logicon211: We don't. I'll remove it

README.md
```diff
@@ -10,12 +10,47 @@ python stix_shifter.py `<data source>` `<input format>` `<STIX pattern>`
...
+```json
+[
+  {
+    "x_com_ibm_uds_datasource": {
+      "id": "7c0de425-33bf-46be-9e38-e42319e36d95",
+      "name": "events"
+    },
+    "id": "observed-data--600ddcf4-8d21-4667-955a-2d14e157b1f2",
+    "type": "observed-data",
+    "objects": {
```
jleemur: format looks more like:

```
          {
                "created": "2018-05-31T16:26:00.320Z",
                "first_observed": "2018-05-31T16:26:00.320Z",
                "id": "observed-data--8a3e5b82-6a37-4a49-9d3e-3b4608214134",
                "last_observed": "2018-05-31T16:26:00.320Z",
                "modified": "2018-05-31T16:26:00.320Z",
                "number_observed": 1,
                "objects": {
                    "0": {
                        "type": "ipv4-addr",
                        "value": "127.0.0.1"
                    },
                    "1": {
                        "type": "ipv4-addr",
                        "value": "9.21.123.112"
                    },
                    "2": {
                        "dst_ref": "0",
                        "src_ref": "1",
                        "type": "network-traffic"
                    }
                },
                "type": "observed-data",
                "x_com_ibm_ariel": {
                    "category_id": "8052",
                    "category_name": "Information",
                    "identity_ip": "0.0.0.0",
                    "log_source_id": "69",
                    "log_source_name": "Health Metrics-2 :: alex112",
                    "magnitude": "3",
                    "protocol_id": "255",
                    "protocol_name": "Reserved",
                    "qid": "94000001",
                    "qid_name": "Health Metric"
                },
                "x_com_ibm_uds_datasource": {
                    "id": "7c0de425-33bf-46be-9e38-e42319e36d95",
                    "name": "events"
                }
            }
```

requirements.txt
```diff
@@ -17,4 +17,6 @@ pytest==3.5.0
 pytest-cov==2.5.1
 six==1.11.0
 stix-pattern-translator==0.1.0
+stix2-patterns==0.6.0
```
jleemur: general question... why don't we have requirements-dev.txt in this project?

README.md
```diff
@@ -10,12 +10,47 @@ python stix_shifter.py `<data source>` `<input format>` `<STIX pattern>`
 
 Example of converting a STIX pattern to an AQL query:
 
-`python3 stix_shifter.py "qradar" "sco" "[ipv4-addr:value = '192.168.122.83' or ipv4-addr:value = '192.168.122.84']"`
+`python stix_shifter.py "qradar" "sco" "[ipv4-addr:value = '192.168.122.83' or ipv4-addr:value = '192.168.122.84']"`
```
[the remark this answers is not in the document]
delliott90: Python 3 is still required, but I would think most people would have the python command pointing to python 3 and so that's why I suggested it be removed from the readme. Maybe I'm wrong on that.

requirements.txt
Logicon211: We didn't really get that far. The standard convention is to just have a requirements.txt. Travis actually looks for requirements.txt by default so in our private repo's we've configured it to use requirements-dev instead.

We should probably make one though

PR #6, opened 2018-06-01
Add command line parser to Stix Shifter

benjamin-craig:


benjamin-craig:
This will have to change most likely because of Logan's PR.

Logicon211:
There's a decent amount that's different. We should probably talk about possible inputs.

benjamin-craig:
We will merge your stuff first then I will fix this one up.

PR #8, opened 2018-06-04
add aql_passthrough module

Logicon211:


src/modules/aql_passthrough/json/to_stix_map.json
```diff
@@ -0,0 +1,123 @@
+{
```
delliott90: Unless something has changed in this mapping, I'd just reference the one in the QRadar module rather than repeating code. I'm not sure we should even have aql_passthrough as a separate module since it's still translating from QRadar. Wouldn't it be better to use the QRadar module but skip the stix-to-aql translation?

Logicon211: I talked with Ben and Misha about this. Currently the only way to actually change which module you're using is to give it a different module name. So in order to use aql passthrough it needs to be a separate module option.

Because multiple modules might be using the JSONToSTIX result translator, we also need to make sure they're able to define a default mapping file. For example if someone made a splunk module and they used JSONToSTIX as the result translator then they would most certainly have a different default mapping. So to keep it a bit better organized, each module should house their own default mapping files

delliott90:
Given Logan's comments, I'm good with merging this. We can always combined the AQL passthrough and QRadar modules if we need to.

stix_shifter.py
```diff
@@ -2,7 +2,7 @@
 import sys
 import importlib
 
-MODULES = ['qradar', 'dummy']
+MODULES = ['qradar', 'dummy', 'aql_passthrough']
```
mdazam1942: this PR contains the same change: [https://github.com/IBM/stix-shifter/pull/9](url)
Should I ask Ben to discard his change?

PR #10, opened 2018-06-04
More stix to aql work

delliott90:
- Added mapping for STIX file and network-traffic objects. Still need to figure out how to handle queries for file hashes.
- Added parenthesis to resulting AQL so that multiple QRadar fields that map to a STIX type can be group as OR checks.

### DIFF
--- a/src/modules/qradar/aql_query_constructor.py
+++ b/src/modules/qradar/aql_query_constructor.py
@@ -99,12 +99,20 @@ class AqlQueryStringPatternTranslator:
             comparison_string = ""
             mapped_fields_count = len(mapped_fields_array)
 
+            group_query = (mapped_fields_count > 1)
+
+            if(group_query):
+                comparison_string += "("
+
             for mapped_field in mapped_fields_array:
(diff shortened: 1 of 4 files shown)


src/modules/qradar/aql_query_constructor.py
```diff
@@ -99,13 +99,21 @@ def _parse_expression(self, expression) -> str:
...
             for mapped_field in mapped_fields_array:
                 comparison_string += "{mapped_field}{comparator}{value}".format(
                     mapped_field=mapped_field, comparator=comparator, value=value)
                 if (mapped_fields_count > 1):
                     comparison_string += " OR "
                     mapped_fields_count -= 1
 
+            if(group_query):
+                comparison_string += ")"
+
```
mdazam1942: Would this below code work? Just a suggestion, cause i see you had to check twice and add open and close parentheses in two different places.

```
if(group_query):
             comparison_string = "(" + comparison_string + ")"
```

delliott90: I had to use a second variable otherwise the string would be repeated twice, but yes that does read better than what I had.

PR #17, opened 2018-06-18
feat(#310) refactor json_to_stix and reference methods

rcasey92:
Updating the transform method, and updating mapping/network-traffic object.


stix_shifter/src/json_to_stix/json_to_stix.py
```diff
@@ -1,4 +1,5 @@
 import json
+import os
```
delliott90: I think you forgot to remove this.

PR #24, opened 2018-07-04
fix: refactored transform logic

rcasey92:


delliott90:
Artifact objects are no longer going into the objects list.
**Before**
`"2": { "payload_bin": "payloadGoesHere", "type": "artifact" },`
**After**
`"artifact": { "payload_bin": "payloadGoesHere" },`

stix_shifter/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -17,184 +17,152 @@ def convert_to_stix(datasource, map_data, data, transformers, options):
...
-        self.observation_props = observable.observation_props
+    @staticmethod
+    def _split_key(key, type_name=None):
+        try:
+            if type_name:
+                return key.split('.') if key.index('.') else None
+            else:
+                return key.replace('-', '_').split('.') if key.index('.') else None
+        except ValueError:
+            print('{0} cannot be split'.format(key))
```
delliott90: I'm not sure it's good practice to trigger an error here given that we expect this to happen for several of our input attributes (such as last_observed). I feel errors should be reserved for when things actually go wrong. It may be better to bring the logger back in so you can use the logging.debug call.

delliott90: You may also wish to keep the comment about why dashes are being replaced by underscores. You could just add this to the docstring assuming you go back to add some.

tests/json_to_stix/test_class.py
```diff
@@ -48,19 +48,31 @@ def test_observation_prop(self):
...
-            "type": "value"
-        }, "payload": {
-            "key": "artifact.payload_bin",
-            "type": "value"
+        datasource = {
+            'id': '123',
+            'name': 'sourcename'
         }
+        map_data = {
+            "ip": {
```
delliott90: I know this was previously in the test, and won't even matter if we just use the existing map file for testing, but ip doesn't actually come back from QRadar (or at least it's not in our mapping). This is pretty much handled by the source and destination ip tests anyway.

PR #25, opened 2018-07-11
Identity object support

delliott90:
- STIX results are now wrapped in a bundle
- An identity object representing a data source is now passed into stix-shifter and added to the bundle. - Refactored AQL Json to STIX tests


stix_shifter/src/json_to_stix/json_to_stix.py
```diff
@@ -28,11 +29,7 @@ def translate_results(self, data, options, mapping=None):
...
-        # todo: make datasource id/name dynamic
-        datasource = {
-            'id': '7c0de425-33bf-46be-9e38-e42319e36d95', 'name': 'events'}
-
-        results = json_to_stix_translator.convert_to_stix(datasource, map_data,
+        results = json_to_stix_translator.convert_to_stix(data_source, map_data,
                                                           json_data, transformers.get_all_transformers(), options)
 
-        return json.dumps(results, indent=4, sort_keys=True)
+        return json.dumps(results, indent=4, sort_keys=False)
```
rcasey92: Why change sort_keys to false? slight performance increase?

delliott90: It was just so the bundle name and id show up at the top. The sorting in this can now effects the bundle rather than the elements in the observed data.

PR #27, opened 2018-07-26
feat(car-json-to-stix): Added new module for translating CAR objects to STIX

samcornwell:
Clean new module with tests and docs. Does not currently include the full CAR data model (https://car.mitre.org/wiki/Data_Model), but should work with most `process` and `flow` (`network-traffic`) CAR objects.


rcasey92:
I think that the following stix properties are still missing from the mapping file:

screen shot 2018-08-02 at 4 16 25 pm

PR #35, opened 2018-09-10
feat: merge stix-pattern-translator codebase into stix-shifter

samcornwell:
Merge stix-pattern-translator core into stix-shifter and create/extend splunk, elastic, cim, and car modules.


delliott90:
Tested the translation from STIX patterns to Splunk and Elastic search queries and the results match with what I saw with the translation project. Everything still works. Good to merge.

PR #43, opened 2018-09-17
Feature splunk issuperset & issubset operator

bhaveshkumarraj:
It has ISSUPERSET and ISSUBSET operator implementation. Also, there are test cases for Splunk translator.


HedleyWard:
Rebased master into branch and made a correction in Splunk to SPL test cases which was causing the Travis build to fail.

PR #49, opened 2018-09-24
WIP CloudSQL Module

imolloy:
For log data stored in cloud object store (COS) and queried using CloudSQL.


stix_shifter/src/modules/csa/json/at_event_fields.json
```diff
@@ -0,0 +1,29 @@
...
+    "initiator.credential.type as initiator_credential_type",
+   	"ALCH_ACCOUNT_ID as alch_account_id",
+   	"ALCH_TENANT_ID as alch_tenant_id",
+   	"eventTime as eventTime",
+   	"action as action",
+    "target.id as target_id",
+    "target.name as target_name",
+    "event_uuid as event_uuid"
+  ],
+  "domain-name": [
```
delliott90: Unless this is used by Cloud SQL, you can probably get rid of the domain-name section here. We removed it from the QRadar module since it was never being used.

stix_shifter/src/modules/csa/json/to_stix_map.json
```diff
@@ -0,0 +1,116 @@
+{
```
delliott90: Does this mapping work with CloudSQL? This looks like the same mapping used for QRadar.

PR #52, opened 2018-09-26
Millisecond timestamp support

delliott90:
- Fix for milliseconds in START STOP timestamps
- Removed domain name from Qradar mappings since it's not a supported AQL field
- Added aql-to-stix mapping for filename. We had this mapped from six-to-aql, but was never handling it coming back from the datasource.

Note that like URL, Filename is a custom AQL attribute so we may want think about removing these from the default mapping files.


delliott90:
Fixes https://github.com/IBM/stix-shifter/issues/46

PR #57, opened 2018-10-05
Custom mappings

delliott90:
Six-shifter now takes in an options hash that allows for custom translation mappings for STIX patterns to AQL query, custom AQL field select, custom JSON to STIX mapping (for data sources that return JSON), custom query limit, and custom time window (when START STOP aren't used).

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,8 +7,15 @@ import json
 def __main__():
     """
     In the case of converting a stix pattern to datasource query, arguments will take the form of...
-    <module> <translate_type> <data>
+    <module> <translate_type> <data> <options>
     The module and translate_type will determine what module and method gets called
+    Options argument comes in as:
+      "{
+          "select_fields": <string array of fields in the datasource select statement>},
+          "mapping": <mapping hash for either stix pattern to datasource or data results to stix observation objects>,
+          "result_limit": <integer to limit number or results in the data source query>,
+          "time_range": <time window (ie. last 5 minutes) used in the data source query when START STOP qualifiers are absent>
+       }"
     """
 
     # process arguments
@@ -33,7 +40,6 @@ def __main__():
                                   help='run stix2 validator against the converted results')
     translate_parser.add_argument('-m', '--data-mapper',
                                   help='module to use for the data mapper')
-    # translate_parser.add_argument('-options', help='options that can be passed in')
 
     args = parser.parse_args()
 
@@ -41,7 +47,6 @@ def __main__():
         parser.print_help(sys.stderr)
         sys.exit(1)
 
-    # args.options = "{\"select_fields\": [\"<FIELDS YOU WANT TO BRING BACK>\"}, \"from_stix_map\": {<HASH OF MAPPING>}}"
     options = json.loads(args.options) if bool(args.options) else {}
     if args.stix_validator:
         options['stix_validator'] = args.stix_validator
--- a/main.py
+++ b/main.py
@@ -1,6 +1,7 @@
 import argparse
 import sys
 from stix_shifter import stix_shifter
+import json
 
 
 def __main__():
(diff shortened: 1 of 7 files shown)


main.py
```diff
@@ -1,13 +1,21 @@
...
     In the case of converting a stix pattern to datasource query, arguments will take the form of...
-    <module> <translate_type> <data>
+    <module> <translate_type> <data> <options>
     The module and translate_type will determine what module and method gets called
+    Options argument comes in as:
+      "{
+          "select_fields": <string array of fields in the datasource select statement>},
+          "mapping": <mapping hash for either stix pattern to datasource or data results to stix observation objects>,
+          "result_limit": <integer to limit number or results in the data source query>,
+          "time_range": <time window (ie. last 5 minutes) used in the data source query when START STOP qualifiers are absent>
```
jleemur: What are you accepting for a time_range value? An integer?

stix_shifter/src/modules/qradar/stix_to_aql.py
```diff
@@ -24,7 +24,9 @@ def transform_query(self, data, options, mapping=None):
         logger.info("Converting STIX2 Pattern to ariel")
 
         query_object = generate_query(data)
-        data_model_mapper = qradar_data_mapping.QRadarDataMapper()
+        data_model_mapper = qradar_data_mapping.QRadarDataMapper(options)
+        result_limit = options['result_limit'] if 'result_limit' in options else 'limit 10000'
+        timerange = options['timerange'] if 'timerange' in options else None
```
jleemur: In docs you have `time_range`

main.py
delliott90: No, a string. The assumption is that when you are translating to the datasource query, you know what format the time range should look like. Hence the example of "last 5 minutes"

stix_shifter/src/modules/qradar/stix_to_aql.py
delliott90: I updated the doc to read "timerange".

stix_shifter/src/modules/qradar/aql_query_constructor.py
```diff
@@ -205,10 +206,19 @@ def parse_expression(self, pattern: Pattern):
...
-def translate_pattern(pattern: Pattern, data_model_mapping):
-    x = AqlQueryStringPatternTranslator(pattern, data_model_mapping)
+def _test_for_start_stop(query_string) -> bool:
+    pattern = "START'\d{4}(-\d{2}){2}\s\d{2}(:\d{2}){2}(\.\d+)?Z?'\s?STOP"
+    match = re.search(pattern, query_string)
+    return bool(match)
+
+
+def translate_pattern(pattern: Pattern, data_model_mapping, result_limit, timerange=None):
+    x = AqlQueryStringPatternTranslator(pattern, data_model_mapping, result_limit)
```
rcasey92: might be better to have a more descriptive variable name

stix_shifter/src/json_to_stix/json_to_stix.py
```diff
@@ -19,15 +19,17 @@ def translate_results(self, data_source, data, options, mapping=None):
         :return: STIX formatted results
         :rtype: str
         """
+
+        self.mapping_json = options['mapping'] if 'mapping' in options else {}
         json_data = json.loads(data)
         data_source = json.loads(data_source)
 
-        if(mapping is None):
+        if(not bool(self.mapping_json)):
```
rcasey92: Why use : `if not bool(self.mapping_json):`

instead of :`if not self.mapping_json:`?

Is there an edge case that it covers?

rcasey92:
Looks good to me other than those little syntax issues

PR #58, opened 2018-10-09
fix: add optional timerange to AQL query

delliott90:
`result_limit` and `timerange` are passed in as integer options. Result limit defaults to 10000 and time range defaults to 5 (minutes)

### DIFF
--- a/stix_shifter/src/modules/qradar/aql_query_constructor.py
+++ b/stix_shifter/src/modules/qradar/aql_query_constructor.py
@@ -168,7 +168,7 @@ class AqlQueryStringPatternTranslator:
             if expression.negated:
                 comparison_string = self._negate_comparison(comparison_string)
             if qualifier is not None:
-                return "SPLIT{} {} {}SPLIT".format(comparison_string, self.result_limit, qualifier)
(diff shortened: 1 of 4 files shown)


stix_shifter/src/modules/qradar/aql_query_constructor.py
```diff
@@ -217,8 +217,13 @@ def translate_pattern(pattern: Pattern, data_model_mapping, result_limit, timera
...
     for where_statement in translated_where_statements.queries:
-        if(_test_for_start_stop(where_statement)):
-            queries.append("SELECT {} FROM events WHERE {}".format(select_statement, where_statement))
+        has_start_stop = _test_for_start_stop(where_statement)
+        if(has_start_stop):
+            query = "SELECT {} FROM events WHERE {}".format(select_statement, where_statement)
         else:
-            queries.append("SELECT {} FROM events WHERE {} {}".format(select_statement, where_statement, result_limit))
+            query = "SELECT {} FROM events WHERE {} limit {}".format(select_statement, where_statement, result_limit)
+        if(not has_start_stop):
```
rcasey92: couldn't this be added to the above else statement?

delliott90: Yes, I don't know what I was thinking.

PR #59, opened 2018-10-11
Custom mapping splunk

bhaveshkumarraj:
Implemented result limit and time range functionalities. Also added custom mapping support.


stix_shifter/src/modules/splunk/splunk_query_constructor.py
```diff
@@ -117,7 +123,9 @@ class _ObservationExpressionTranslator:
         ComparisonComparators.In: encoders.set,
         ComparisonComparators.Matches: encoders.matches,
         ComparisonExpressionOperators.And: 'AND',
-        ComparisonExpressionOperators.Or: 'OR'
+        ComparisonExpressionOperators.Or: 'OR',
+        ComparisonComparators.IsSuperSet: "=", # Splunk cidrmatch('<ip in CIDR notation>', ip) and = operator give the same result
+        ComparisonComparators.IsSubSet: "=" # cidrmatch function can be used for both issuperset and issubset operator
```
mdazam1942: may be my lack of knowledge, how does superset and subset works with the same operator? I don't see any difference in the translation. the splunk query translation returns the same result in both cases.

stix_shifter/src/modules/splunk/README.md
```diff
@@ -73,4 +73,4 @@ Will return the following valid STIX Cyber Observable Object:
         }                                                                                               
     ]                                                                                                   
 }                                                                                                       
-```
\ No newline at end of file
+```
```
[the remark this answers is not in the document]
bhaveshkumarraj: That's fine, it is just end of JSON markup.

stix_shifter/src/patterns/pattern_objects.py
```diff
@@ -14,9 +14,8 @@ class ComparisonComparators(Enum):
         In,
         Like,
         Matches,
-        IsSubSet,
-        IsSuperSet
```
delliott90: I don't think this needs to be removed as some other datasource may be able to eventually use the SuperSet operator. You've already written the exitPropTestIsSuperset in the parser, so may as well leave this here.

stix_shifter/src/stix_pattern_parser/stix_pattern_parser.py
```diff
@@ -20,8 +20,7 @@ class PatternTranslator:
         # Treat AND's as OR's -- Unsure how two ObsExps wouldn't cancel each other out.
         ObservationOperators.And: 'OR',
         ObservationOperators.FollowedBy: 'FOLLOWEDBY',
-        ComparisonComparators.IsSubSet: 'ISSUBSET',
-        ComparisonComparators.IsSuperSet: 'ISSUPERSET',
```
delliott90: Same as the above comment. Best to leave SuperSet operator in even though nothing is calling it right now.

tests/splunk_json_to_stix/test_class.py
```diff
@@ -362,4 +368,54 @@ def test_email_cim_to_stix(self):
...
         assert(addr_obj['value'] == src_user)
-        
\ No newline at end of file
+    
+    def test_custom_mapping(self):
+
+        data_source = "{\"type\": \"identity\", \"id\": \"identity--3532c56d-ea72-48be-a2ad-1a53f4c9c6d3\", \"name\": \"Splunk\", \"identity_class\": \"events\"}"
+        data = "[{\"tag\":\"network\", \"src_ip\": \"127.0.0.1\"}]"
+
+        options = {
```
delliott90: Judging from json_to_stix.py, wouldn't this need to come in as options = {'mapping': {<mapping json>}} ?

bhaveshkumarraj: we're using cim_to_stix.py which is inside Splunk module folder, however, I have changed it to take from mapping object inside options.

PR #63, opened 2018-10-30
Transform START STOP timestamps to epoch time

delliott90:


stix_shifter/src/transformers.py
```diff
@@ -43,8 +45,7 @@ class TimestampToEpoch(ValueTransformer):
     def transform(timestamp):
         time_pattern = '%Y-%m-%dT%H:%M:%S.%fZ'
         epoch = datetime(1970, 1, 1)
-        converted_epoch = int(
-            (datetime.strptime(timestamp, time_pattern) - epoch).total_seconds())
+        converted_epoch = int(((datetime.strptime(timestamp, time_pattern) - epoch).total_seconds()) * 1000)
```
mdazam1942: are converting to epoch/unix or milliseconds? epoch is 10 digits and milliseconds is 13

PR #64, opened 2018-10-30
Add domain-name support for stix patterning and translation

mdazam1942:


stix_shifter/src/transformers.py
```diff
@@ -116,6 +117,19 @@ def transform(obj):
...
             print("Cannot convert input to file name")
 
+class ToDomainName(ValueTransformer):
+    """A value transformer for expected domain name"""
+
+    @staticmethod
+    def transform(url):
+        try:
+            parsed_url = urlparse(url)
+            host_name = parsed_url.netloc.split(".")
```
delliott90: You may be able to just return parsed_url.netloc

PR #68, opened 2018-11-14
Qradar filehash

delliott90:
Adds file hash support to the QRadar module. In cases where QRadar returns an ambiguous hash type, the log source id is examined to attempt to determine the type


stix_shifter/src/modules/qradar/json/to_stix_map.json
```diff
@@ -95,7 +95,24 @@
...
+  "sha1hash": {
+    "key": "file.hashes.SHA-1",
+    "object": "fl"
+  },
+  "md5hash": {
+    "key": "file.hashes.MD5",
+    "object": "fl"
+  },
+  "filehash": {
+    "key": "file.hashes.UNKNOWN",
```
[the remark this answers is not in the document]
delliott90: No, and it should never be stored in the STIX bundle as that either. I only have it there because the generic "filehash" column still needs to be captured and parsed into STIX. Basically, this mapping gets overridden when the filehash attribute gets put into STIX.

stix_shifter/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -4,6 +4,7 @@
 
 from . import observable
 from stix2validator import validate_instance, print_results
+from stix_shifter.src.modules.qradar import qradar_utils
```
emishas: json to Stix translator not supposed to import QRadar or any other module,
that's why I was suggesting doing it via a callback

stix_shifter/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -138,20 +140,35 @@ def transform(self, obj):
...
         for ds_key in obj:
             if ds_key not in ds_map:
                 logging.debug('{} is not found in map, skipping'.format(ds_key))
                 continue
+
+            generic_hash_key = ''
+            # Check and handle file hash of unknown type
+            if self.hash_options:
+                try:
+                    generic_hash_key = qradar_utils.lookup_hash_with_logsource_id(
```
emishas: the callback idea is that `def transform(self, obj):` would become `def transform(self, obj, callback):`
and then
```
generic_hash_key = callback(...)
```

PR #69, opened 2018-11-16
Dummy module update

delliott90:
Added mapping and logic to the dummy module so that it actually does something now. Ideally, this module will be a good starting point for anyone wanting to add a new data source module.


README.md
```diff
@@ -27,9 +39,16 @@ You may want to use this library and/or contribute to development, if any of the
...
 
-### Call the stix_shifter in the format of
+## Translation
+
+### Converting from STIX Patterns to data source queries (query) or from data source results to STIX cyber observables (results)
+
+#### Call the stix_shifter in the format of
 
 ```
 usage: stix_shifter.py translate [-h]
```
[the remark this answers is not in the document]
delliott90: The examples below are using the CLI, so main.py is what's called. Main.py would never be touched when importing STIX-shifter into another project.

stix_shifter/src/modules/dummy/dummy_data_mapping.py
```diff
@@ -0,0 +1,31 @@
...
+        return {}
+
+
+class DummyDataMapper:
+
+    def __init__(self, options):
+        self.mapping_json = options['mapping'] if 'mapping' in options else {}
+
+    def map_field(self, stix_object_name, stix_property_name):
+        self.map_data = self.mapping_json or _fetch_mapping()
```
emishas: why not call `_fetch_mapping()` inside the constructor instead?
if `self.mapping_json` is not set then you call  `_fetch_mapping()` every call of `map_field `

besides, `self.mapping_json` will never ever be not set now as you set it to empty object `{}`

README.md
```diff
@@ -1,5 +1,17 @@
...
 
+#### Table of Contents
+
+[Introduction](#introduction)<br/>
+[How to use](#How-to-use)<br/>
+[Translation](#translation)<br/>
+[Transmission](#transmission)<br/>
+[Contributing](#contributing)<br/>
+[Creating new modules](#creating-new-modules)<br/>
+[Licencing](#licensing)
```
jleemur: Typo: Licensing

README.md
```diff
@@ -71,6 +194,58 @@ optional arguments:
   -h, --help            show this help message and exit
 ```
 
+### Transmission is called with the following ordered parameters:
+
+```
+<Data Source (ie. "qradar")> <Connection Params: '{"host":"host ip address", "port":"port number", "cert":"certificate"}'> <'{"auth": {authentication}}'> <Transmission Operation: ping, query, status, results or is_async> <Operation input>
+```
+
+**Data source:** This will be the name of the module used for transmission. Currently stix-shifter supports QRadar, Splunk, and Bigfix.
```
jleemur: Translation supports: `QRadar, Splunk, and Elastic`
Transmission supports: `QRadar, Splunk, BigFix`

Making sure this is right

delliott90: Yes that's correct. Bigfix translation is still in progress but its transmission module is in. Elastic isn't an actually data source type so it has no transmission module.

stix_shifter/src/modules/dummy/stix_to_dummy_query.py
```diff
@@ -0,0 +1,31 @@
+import logging
```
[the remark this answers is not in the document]
delliott90: It's used in the transform_query method. It may be an oversight that the log message isn't in the Splunk module. I'll add it in.

PR #72, opened 2018-11-30
Restructure project

jleemur:


jleemur:
I needed to change travis CI settings for the build to pass...

Changed src folder from `stix_shifter/src` to `stix_shifter/`

main.py
```diff
@@ -1,6 +1,6 @@
 import argparse
 import sys
-from stix_shifter import stix_shifter
+from stix_shifter.stix_translation import stix_translation
```
delliott90: Could we move the transmit method into stix_transmission.py? You could then add 
`from stix_shifter.stix_transmission import stix_transmission` here. Right now when I call transmit from the CLI, I get an error saying no module named 'stix_transmission'.

jleemur:
Out of date, new PR is https://github.com/IBM/stix-shifter/pull/76

PR #74, opened 2018-12-10
Fixing request library for Qradar, Bigfix, and Splunk to handle certificates

benjamin-craig:


jleemur:
Just waiting to test QRoc

PR #75, opened 2018-12-10
Fix: qroc proxy support back port

benjamin-craig:
@emishas Asked misha for help testing the port.  He will send me information when he can.


jleemur:
Rebased PR, and fixed transmit for qroc. Looks good to me

benjamin-craig:
I checked it out and tested it.

PR #81, opened 2019-01-02
BigFIx module update for translations

mdazam1942:
This part of the bf module update. We need to review the whole module. We merged the previous update without CR.


stix_shifter/stix_translation/src/modules/bigfix/bigfix_query_constructor.py
```diff
@@ -8,48 +8,33 @@ class RelevanceQueryStringPatternTranslator:
     comparator_lookup = {
         ComparisonExpressionOperators.And: "AND",
         ComparisonExpressionOperators.Or: "OR",
-        ComparisonComparators.GreaterThan: ">",
-        ComparisonComparators.GreaterThanOrEqual: ">=",
-        ComparisonComparators.LessThan: "<",
-        ComparisonComparators.LessThanOrEqual: "<=",
+        # ComparisonComparators.GreaterThan: ">",
```
delliott90: Unless you think we will support these later, it's safe to delete the commented code. It can always be added from the dummy module later.

stix_shifter/stix_translation/src/modules/bigfix/bigfix_to_stix/bigfix_to_stix_translator.py
```diff
@@ -0,0 +1,214 @@
+import re
```
delliott90: We probably want to avoid copying the entire json_to_stix_translator file.

stix_shifter/stix_translation/src/modules/bigfix/bigfix_to_stix/observable.py
```diff
@@ -0,0 +1,26 @@
+REGEX = {
```
delliott90: This file already exits in json_to_stix so it's best to use that rather than repeat code.

stix_shifter/stix_translation/src/modules/bigfix/bigfix_to_stix/bigfix_to_stix_translator.py
mdazam1942: there are lot of changes i had to make inside this file. i also had to reformat the json. didn't want to break the base translator at the first place. you can have a look at this translator for now. i'm working on to make it compatible with base  json_to_stix_translator if i can avoid copying the code i will push new changes

stix_shifter/stix_translation/src/modules/bigfix/bigfix_to_stix/observable.py
mdazam1942: yup. i will use that for sure.

stix_shifter/stix_translation/src/modules/bigfix/bigfix_query_constructor.py
mdazam1942: sure.

stix_shifter/stix_translation/src/modules/bigfix/bigfix_to_stix/bigfix_to_stix_translator.py
delliott90: Did you look at what I did for the QRadar module? I defined a class and method in the qradar_utils.py file, passed that into the JSONToStix call in qradar_translator (this would be bixfix_translator in your case) and then use that passed-in class to handle any QRadar specific logic.

mdazam1942: so i haven't created any util class but formatted the output in transmission `bigfix_results_connector.py` so the translation interface will get well formatted json so that we can use our base json_to_stix classes.

requirements-dev.txt
```diff
@@ -18,4 +18,5 @@ pytest-cov==2.5.1
 six==1.11.0
 wrapt==1.10.11
 requests==2.20.1
-flask==1.0.2
\ No newline at end of file
+flask==1.0.2
+pyyaml==3.13
```
delliott90: Is pyyaml still needed? I can't find where it's used.

mdazam1942: ohh forgot to take it out. yeah dont need it anymore

stix_shifter/stix_translation/src/modules/bigfix/bigfix_query_constructor.py
```diff
@@ -69,6 +48,7 @@ def _parse_expression(self, expression, qualifier=None):
             stix_object, stix_remainings = expression.object_path.split(':')
             value = self._escape_value(expression.value)
             comparator = self.comparator_lookup[expression.comparator]
+            final_value = comparator + ' "' + value + '"'
             stix_property = ""
             stix_key = ""
             self.query_type = stix_object
```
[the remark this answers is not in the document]
mdazam1942: single hash query can also be interpreted as file hash query not just process. if file hash passed into stix pattern, it can also be someone looking for files with that hash. not sure how are going to differentiate that?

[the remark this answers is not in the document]
benjamin-craig: Well if we just have a file hash we do not have any directories so I think processes works better

stix_shifter/stix_translation/src/modules/bigfix/bigfix_query_constructor.py
```diff
@@ -8,48 +8,27 @@ class RelevanceQueryStringPatternTranslator:
...
-        "process_name_hash": "( name of it | \"n/a\", process id of it as string | \"n/a\", \"sha256\", sha256 of image file of it | \"n/a\", pathname of image file of it | \"n/a\" ) of processes whose ({stix_object1} of it as lowercase contains \"{expr1_value}\" as lowercase AND {stix_object2} of image file of it as lowercase = \"{expr2_value}\" as lowercase )"
+        "all_files_in_directory": "(\"file\", name of it | \"n/a\", \"sha256\", sha256 of it | \"n/a\", \"sha1\", sha1 of it | \"n/a\", \"md5\", md5 of it | \"n/a\", pathname of it | \"n/a\", (modification time of it - \"01 Jan 1970 00:00:00 +0000\" as time)/second ) of files of folder (\"{file_path}\")",        
+        "file_query": "(\"file\", name of it | \"n/a\", \"sha256\", sha256 of it | \"n/a\", \"sha1\", sha1 of it | \"n/a\", \"md5\", md5 of it | \"n/a\", pathname of it | \"n/a\", (modification time of it - \"01 Jan 1970 00:00:00 +0000\" as time)/second ) of files whose ({stix_object} of it as lowercase {object_value} as lowercase) of folder (\"{file_path}\")",
```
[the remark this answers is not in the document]
delliott90: Yeah, you are correct. file:hashes.sha256 wouldn't be valid STIX, it needs to be with the 'SHA-256', 'SHA-1', or 'MD5'

PR #97, opened 2019-01-17
splunk transmission updates

benjamin-craig:


stix_shifter/stix_transmission/src/modules/splunk/spl_api_client.py
```diff
@@ -75,7 +75,7 @@ def get_search_results(self, search_id, offset, count):
             data['offset'] = str(offset)
             data['count'] = str(count)
         # response object body should contain information pertaining to search.
-        return self.client.call_api(endpoint, 'GET', data=data)
```
yurii-klymenko: both ways work to me, they just fetch different amount of data.
The question is which way is correct?
for query: 'SELECT QIDNAME(qid) as qidname, qid as qid, CATEGORYNAME(category) as categoryname, category as categoryid, CATEGORYNAME(highlevelcategory) as high_level_category_name, highlevelcategory as high_level_category_id, logsourceid as logsourceid, LOGSOURCETYPENAME(logsourceid) as logsourcetypename, LOGSOURCENAME(logsourceid) as logsourcename, starttime as starttime, endtime as endtime, devicetime as devicetime, sourceaddress as sourceip, sourceport as sourceport, sourcemac as sourcemac, destinationaddress as destinationip, destinationport as destinationport, destinationmac as destinationmac, username as username, eventdirection as direction, identityip as identityip, identityhostname as identity_host_name, eventcount as eventcount, PROTOCOLNAME(protocolid) as protocol, BASE64(payload) as payload, URL as url, magnitude as magnitude, Filename as filename, URL as domainname FROM events WHERE url = '"'"'www.example.com'"'"' limit 10000 last 5 minutes'
parameters in body (before fix) way returns 94 data items;
parameters in url (after fix) way returns 10 data items.

benjamin-craig: the limit is 10000 (this means its a total of 10000 results in the search id cursor)
but when we go to fetch the 10000 items we want to fetch them in sizes of 1000 at a time.  So we pass 0 to 1000 for first 1000, then we do 1001 to 2000 to get the next 1000.

tests/stix_transmission/splunk/test_class.py
```diff
@@ -83,7 +107,33 @@ def test_query_response(self, mock_query_response, mock_api_client):
...
         assert 'search_id' in query_response
         assert query_response['search_id'] == "1536672851.4012"
- 
+
+    @patch('stix_shifter.stix_transmission.src.modules.splunk.spl_api_client.APIClient.create_search')
+    def test_query_response_exception(self, mock_query_response, mock_api_client):
+        mock_api_client.return_value = None
+        mocked_return_value = '{"sid":"1536672851.4012"}'
+        mock_query_response.return_value = SplunkMockResponse(201, mocked_return_value)
+        mock_query_response.side_effect = Exception('an error occurred creating query')
```
yurii-klymenko: the text is a little bit confusing, the text is not checked in the test but it looks pretty similar to the text in the checking. I would replace it with some dummy text and checked it as well.

benjamin-craig:
@yurii-klymenko Made changes to the unit tests.

PR #100, opened 2019-01-19
Dynamic mapping for fields for splunk.

benjamin-craig:
Ex: 

```python main.py translate splunk query {} "[ipv4-addr:value = '58.218.66.97']" '{"fields":{"default":["src_port","src_ip"]}, "mapping": {"ipv4-addr":{ "cim_type": "flow","fields": { "value": ["source","dest"]}}}}'```

### DIFF
--- a/stix_shifter/stix_translation/src/modules/splunk/stix_to_splunk.py
+++ b/stix_shifter/stix_translation/src/modules/splunk/stix_to_splunk.py
@@ -4,6 +4,7 @@ import importlib
 from ...patterns.parser import generate_query
 from ..base.base_query_translator import BaseQueryTranslator
 from . import splunk_query_constructor
+from ..cim import cim_data_mapping
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: stix_shifter/stix_translation/src/modules/splunk/stix_to_splunk.py [hunk 1]
```diff
@@ -30,19 +31,24 @@ def transform_query(self, data, options, mapping=None):
...
-        data_mapper_module_name = ''.join(["stix_shifter.stix_translation.src.modules.", data_mapper, ".", data_mapper, "_data_mapping"])
-
-        try:
-            data_mapper_module = importlib.import_module(data_mapper_module_name)
-            data_model_mapper = data_mapper_module.mapper_class(mapping)
-        except ModuleNotFoundError:
-            raise NotImplementedError(f"Module {data_mapper_module_name} not implemented")
-        except AttributeError:
-            raise NotImplementedError(f"Module {data_mapper_module_name} does not implement mapper_class attribute")
+        if data_mapper != 'cim':
```
mdazam1942: we can do something like: 
```
if not data_mapper:
    data_mapper = 'cim'
    data_mapper_module = cim_data_mapping
    data_model_mapper = data_mapper_module.mapper_class(mapping,fields)
else:
  # put the code for if data_mapper != 'cim':
```
this way we can avoid one extra conditional checking.

mdazam1942: let me know if that works

PR #102, opened 2019-01-21
Implement CarbonBlack ping transmission module

Collin-Richards:
I've implemented the ping transmission module. The other transmission modules components are not yet implemented. I'm not sure if you'd rather have smaller PR's for each component or one larger one with all of them implemented. We can discuss that below.

### DIFF
--- a/stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_api_client.py
+++ b/stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_api_client.py
@@ -1,0 +1,24 @@
+from ..utils.RestApiClient import RestApiClient
+
+
+class APIClient():
+
+    PING_ENDPOINT = 'sensor'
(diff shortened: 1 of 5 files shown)


Collin-Richards:
Also as a side note, I noticed that the travis build is running some linting code, but it doesn't appear to actually care if the linting passes. Is that intentional?

delliott90:
It would be better to have the entire module with both translation and transmission filled out before submitting a PR. Regarding the linting code, it may be that something needs fixed on travis.

Collin-Richards:
Ok. I'll keep working on this branch and submit a new request when both translation and transmission are implemented.

delliott90:
Take a look at this PR, https://github.com/IBM/stix-shifter/pull/96, I simplified the transmission modules a bit for the Dummy and QRadar modules. I put all of the methods in the connector class rather than having separate classes for each method.

PR #109, opened 2019-01-29
Need to make sure we return an empty list for our pipeline.

benjamin-craig:
An empty json object will still generate identity information.

### DIFF
--- a/stix_shifter/stix_transmission/src/modules/splunk/splunk_results_connector.py
+++ b/stix_shifter/stix_transmission/src/modules/splunk/splunk_results_connector.py
@@ -12,11 +12,12 @@ class SplunkResultsConnector(BaseResultsConnector):
         try:
             response = self.api_client.get_search_results(search_id, offset, length)
             response_code = response.code
-            response_json = json.load(response) 
+            response_json = json.load(response)
+
             if "results" in response_json:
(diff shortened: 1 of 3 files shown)


stix_shifter/stix_transmission/src/modules/splunk/splunk_results_connector.py
```diff
@@ -12,11 +12,12 @@ def create_results_connection(self, search_id, offset, length):
         try:
             response = self.api_client.get_search_results(search_id, offset, length)
             response_code = response.code
-            response_json = json.load(response) 
+            response_json = json.load(response)
+
             if "results" in response_json:
-                results = [{}] if (response_json['results'] == []) else response_json['results']
+                results = [] if (response_json['results'] == []) else response_json['results']
```
mdazam1942: kinda confusing to me. i thought we already handled that in StixPatternToQuery, right?
And also not sure why do we need this checking here. if the response is non `200` than this checking is useless. if the successful response always returns with `results` json object than we can just set 
`return_obj['data'] = response_json['results']`

benjamin-craig: It is a one line assignment statement.  I never wrote the code I'm just fixing it.  If results is empty we want to return an empty list for our pipeline.   This is result set translation.

PR #113, opened 2019-02-01
Carbonblack Translate and Transmit Module

Collin-Richards:
Along with implementing the translate and transmit modules, there are a couple extra changes that were made:

- Added support for arrays of stix references. An example of it's usage can be seen in the opened_connection_refs.
- Created StartStopQualifier object and attached the start and stop timestamp to it. The existing qualifier object has a single string and the individual module was responsible for doing validation.
- Added module dialect support to the transmission module.


yurii-klymenko:
I see that status connector implementation is missing here: stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_connector.py

Collin-Richards:
I think that's intentional because the carbonblack api is synchronous. The other synchronous module, aws_security_hub, is also missing this. Also if people would like to test the code on a carbon black instance I can give you temporary access to an API key.

stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_connector.py
```diff
@@ -0,0 +1,62 @@
...
+            return_obj['error'] = 'error when pinging data source: {}'.format(err)
+            return return_obj
+
+    def create_query_connection(self, query):
+        return {"success": True, "search_id": query}
+
+    def create_results_connection(self, search_id, offset, length):
+        try:
+            query = search_id
+            response = self.api_client.run_search(query)
```
emishas: is there a way to fetch particular data range? the offset/length parameters

stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_connector.py
```diff
@@ -0,0 +1,62 @@
...
+            response = self.api_client.ping_box()
+            response_code = response.code
+            response_json = json.loads(response.read())
+
+            return_obj = dict()
+
+            if len(response_json) > 0 and 200 <= response_code < 300:
+                return_obj['success'] = True
+            else:
+                return_obj['success'] = False
```
yurii-klymenko: error answer is supposed to have an 'code' field. Please check other modules(qradar,bigfix and splunk) for error reporting. It may be needed to add error_mapper.py to implement proper error mapping.

stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_connector.py
Collin-Richards: Yes. It supports a start and a rows parameter. I'll add that. That also made me realize that the api defaults to returning only 10 rows.

stix_shifter/stix_transmission/src/modules/carbonblack/carbonblack_connector.py
Collin-Richards: Ok. I'll add that as well.

Collin-Richards:
I've addressed the issues mentioned above.

After noticing that the number of results returned were defaulting to 10, I've done some testing with a different limit. Requesting 2000 records takes about 8 seconds with a simple query (process_name:cmd.exe) and an arbitrary offset. Note in the test environment this has around 72,000 total results. I've tried testing with sorting by a couple different fields in case certain fields were indexed in their database, but they all appear to have similar runtime.

stix_shifter/stix_translation/src/modules/carbonblack/README.md
```diff
@@ -0,0 +1,377 @@
...
+  "filtered": {}
+}
+```
+
+Then the results can be translated with the following command: (Note just the json array associated with the results key should be copied into this command)
+
+## Example STIX Output Format
+
+```
+$ python3 main.py translate carbonblack results '{"id": "blah"}' '<copied results from previous query>'
```
delliott90: You may want to indicate that '{"id": "blah"}' should be the identity object that is shown below as the first object in the 'objects' array.

stix_shifter/stix_translation/src/modules/carbonblack/carbonblack_data_mapping.py
```diff
@@ -0,0 +1,32 @@
+from os import path
+import json
+from ...exceptions import DataMappingException
+
+
+class CarbonBlackDataMapper:
+    def __init__(self, options):
+        self.map_data = self._fetch_mappings()
```
delliott90: I know in QRadar we first look for a mapping in `options['mapping']` in case the end-user want to override mappings on the data source. Do you think similar support is needed here? I realize by seeing this that I failed to add anything covering options in our documentation.

stix_shifter/stix_translation/src/modules/carbonblack/carbonblack_query_constructor.py
```diff
@@ -0,0 +1,252 @@
...
+        ObservationOperators.And: 'or', # This is technically wrong. It should be converted to two separate queries, but
+        # current behavior of existing modules treat operator as an OR.
+        # observation operator AND - both sides MUST evaluate to true on different observations to be true
+    }
+
+    def __init__(self, pattern: Pattern, data_model_mapper, result_limit):
+        self.dmm = data_model_mapper
+        self.pattern = pattern
+        self.result_limit = result_limit
+        print(pattern)
```
delliott90: Probably don't need these print statements.

stix_shifter/stix_translation/src/modules/carbonblack/carbonblack_data_mapping.py
Collin-Richards: I can add support for that back in. Due to the binary and process api having separate mappings it seemed weird to me a user overriding them.

stix_shifter/stix_translation/src/modules/carbonblack/carbonblack_query_constructor.py
```diff
@@ -0,0 +1,252 @@
...
+        return "-({})".format(comparison_string)
+
+    @staticmethod
+    def _to_cb_timestamp(ts: str) -> str:
+        stripped = ts[2:-2]
+        if '.' in stripped:
+            stripped = stripped.split('.', 1)[0]
+        return stripped
+
+    def _fomat_start_stop_qualifier(self, expression, dialect, qualifier : StartStopQualifier) -> str:
```
delliott90: typo in "fomat"

stix_shifter/stix_translation/src/patterns/grammar/STIXPatternParser.py
```diff
@@ -1596,6 +1596,12 @@ def TimestampLiteral(self, i:int=None):
         def STOP(self):
             return self.getToken(STIXPatternParser.STOP, 0)
 
+        def start_time(self):
```
delliott90: I don't know if it's a good idea to add code directly to this file since it is generated from the STIXPattern.g4 file. The grammar files get generated by running 
`antlr4 -Dlanguage=Python3 -listener -visitor STIXPattern.g4`
Even though the g4 file is included in this project it's mostly a copy of https://github.com/oasis-open/cti-stix2-json-schemas/blob/master/pattern_grammar/STIXPattern.g4. I say mostly because it appears the g4 file used in the original Mitre translation project was not from that oasis repo. There are differences and a straight copy breaks a bunch of tests.

Collin-Richards: Ah good catch, I didn't see that it was generated. I'll investigate this a bit further. I do think this should be done in a way so each parser doesn't need to duplicate a regex for the timestamp logic though.

stix_shifter/stix_transmission/stix_transmission.py
```diff
@@ -16,11 +16,22 @@ class StixTransmission:
     init_error = None
     
     def __init__(self, module, connection, configuration):
-        # pass
+        if module not in TRANSMISSION_MODULES :
+            raise NotImplementedError
+
+        dialect = None
```
delliott90: If the dialect is specific to Carbon Black, it may be better to push this logic down to the Connector if possible. I guess I'm not 100% sure what the module looks like before it gets split.

delliott90:
I intentionally called a transmission query with a bad query `'{"query": "username:SYSTEM", "dialect": "binary"}'`. When I called the results with the same query, I got the following error:
`{'success': False, 'error': 'Unhandled exception. Check logs for details.', 'code': 500}`. I wonder if we need some better error messaging here. If I force a transmission error with QRadar I get something like:
`{'success': False, 'error': 'Query f9faf4be-f9a7-4756-912c-a63e55c87ce1 does not exist', 'code': 901}
`. The QRadar module has an ErrorMapper class which takes care of this messaging.  I'm not sure what errors come back from the Carbon Black APIs.

Collin-Richards:
Yeah. We should at least be displaying the error from carbon black in that case. They aren't that rich, but it should be better then the current message.

delliott90:
After speaking with Collin, we've decided to address the error messaging in a future PR.

PR #114, opened 2019-02-05
Refactor how we get splunk status response.

benjamin-craig:


stix_shifter/stix_transmission/src/modules/splunk/splunk_status_connector.py
```diff
@@ -20,18 +25,17 @@ def create_status_connection(self, search_id):
...
             progress = math.ceil(content['doneProgress'] * 100)  # convert 0-1.0 scale to <int>0-100
+            status = content['dispatchState']
 
-            if content['isDone'] is True:
+            if status == StatusSplunk.COMPLETED.value:
                 status = Status.COMPLETED.value
-            elif content['isFailed'] is True:
+            elif status == StatusSplunk.ERROR.value:
                 status = Status.ERROR.value
             elif content['isFinalized'] is True:
```
yurii-klymenko: I would make the checking more strict. like:
```
elif status == StatusSplunk.RUNNING.value:
    status = Status.RUNNING.value
    if content['isFinalized'] is True:
        status = Status.CANCELED.value
else:
    status = Status.CANCELED.value
```

benjamin-craig: I'm not sure if that will work since Splunk has a state called FINALIZING in the status but I cannot find evidence if FINALIZING and isFinalized is the same thing.

PR #116, opened 2019-02-13
translation error reporting

yurii-klymenko:
translation does not throw any exceptions. On failure it returns:
{'success': False, 'code': 310, 'error': 'Unable to map property `x-UNreadable-payload:value` into AQL'}


tests/stix_translation/patterns/test_miscellaneous_tests.py
```diff
@@ -5,11 +5,15 @@
...
 
 
 class TestNotSupported(unittest.TestCase):
     """ Make sure that SearchFeatureNotSupportedErrors are raised appropriately. """
     def test_elastic_followedby(self):
         pattern = "[ipv4-addr:value = '198.51.100.5'] FOLLOWEDBY [ipv4-addr:value = '198.51.100.10']"
-        with self.assertRaises(SearchFeatureNotSupportedError):
-            res = translate(pattern, SearchPlatforms.ELASTIC, DataModels.CAR)
+        result = translate(pattern, SearchPlatforms.ELASTIC, DataModels.CAR)
+        assert result['success'] == False
```
delliott90: Isn't this assertion backwards? Shouldn't it be expectation == result?

delliott90: Actually, never mine, I have it backwards.

stix_shifter/utils/error_response.py
```diff
@@ -7,6 +7,12 @@
 import json
 
 class ErrorCode(Enum):
+    TRANSLATION_NOTIMPLEMENTED_MODE = 100
+    TRANSLATION_MODULE_DEFAULT_ERROR =  200
+    TRANSLATION_MAPPING_ERROR = 310
+    TRANSLATION_STIX_VALIDATION = 320
```
delliott90: How were these error codes decided? Wouldn't a pattern that fails validation be in the 400 range since it's a poorly formatted request?

yurii-klymenko: these codes are not http codes, mapping to http codes will be done in other places.

delliott90:
I think the error message is wrong when converting from data source results to STIX. I just tried running results with an empty identity hash and am getting
`{'success': False, 'code': 200, 'error': 'Error when converting STIX pattern to data source query'}`
It should instead give some messaging saying it couldn't convert results to STIX.

PR #122, opened 2019-03-11
Recursion fix

delliott90:
Increases the recursion depth limit of python to 10,000 so Antlr doesn't choke on large patterns. Tested with a pattern containing around 2000 ORs and 1000 observatios (square brackets).


yurii-klymenko:
Just wondering, is there way to replace the recursion logic with iteration one?

PR #126, opened 2019-03-26
Handled unsupported attribute maps for STIX to AQL translation (ON HOLD)

delliott90:

Before, when a submitted pattern contained an unmapped STIX attribute, an error was immediately thrown when translating from STIX to AQL queries. Now in certain cases, the unmapped comparison expression is removed, and an AQL query is constructed from the mapped elements. Previously, a query string was built from the ANTLR parsing as it was recursively run through the `_parse_expression()` method in `aql_query_constructor.py`. Now `_parse_expression()` inserts the ANTLR elements (queries, operators, START STOP qualifiers) into an object structure where groupings can be preserved; this is then used to build the final query strings using the `_parse_translated_query_objects()` method

**Pattern examples that will still throw the error:**

Pattern contains only one observation and it contains only one comparison expression and it is unmapped:
`[unmapped:attribute = someValue]`

Pattern contains only one observation and it contains multiple comparison expressions, one of which is unmapped but it is joined to another comparison with an AND operator:
`[unmapped:attribute = someValue AND mapped:attribute = someValue OR mapped:attribute = someValue]`

_**NOTE**: Work could still be done to handle comparisons that are grouped by parenthesis rather than throwing out the entire observation_

**Pattern examples that will produce an AQL query that omits the unmapped comparison:**

`[unmapped:attribute = someValue] OR [mapped:attribute = someValue]`
`[unmapped:attribute = someValue OR mapped:attribute = someValue]`

### DIFF
--- a/stix_shifter/stix_translation/src/modules/qradar/aql_query_constructor.py
+++ b/stix_shifter/stix_translation/src/modules/qradar/aql_query_constructor.py
@@ -6,6 +6,7 @@ from stix_shifter.stix_translation.src.json_to_stix import observable
 import logging
 import json
 import re
+import uuid
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


delliott90:
Hold off on merging, I found a defect.

delliott90:
This didn't work for more complicated patterns. Opened a new PR with a better approach

PR #128, opened 2019-04-03
Support for unmapped STIX attributes in AQL translation

delliott90:

Before, when a submitted pattern contained an unmapped STIX attribute, an error was immediately thrown when translating from STIX to AQL queries. Now in certain cases, the unmapped comparison expression is removed, and an AQL query is constructed from the mapped elements. Previously, a query string was built from the ANTLR parsing as it was recursively run through the `_parse_expression()` method in `aql_query_constructor.py`. Now `_parse_expression()` inserts the ANTLR elements (queries, operators, START STOP qualifiers) into an object structure where groupings can be preserved; this is then used to build the final query strings using the `_parse_translated_query_objects()` method

**Pattern examples that will still throw the error:**

Pattern contains only one observation and it contains only one comparison expression and it is unmapped:
`[unmapped:attribute = someValue]`

Pattern contains only one observation and it contains multiple comparison expressions, one of which is unmapped but it is joined to another comparison with an AND operator:
`[unmapped:attribute = someValue AND mapped:attribute = someValue OR mapped:attribute = someValue]`

_**NOTE**: Work could still be done to handle comparisons that are grouped by parenthesis rather than throwing out the entire observation_

**Pattern examples that will produce an AQL query that omits the unmapped comparison:**

`[unmapped:attribute = someValue] OR [mapped:attribute = someValue]`
`[unmapped:attribute = someValue OR mapped:attribute = someValue]`

### DIFF
--- a/stix_shifter/stix_translation/src/modules/qradar/aql_query_constructor.py
+++ b/stix_shifter/stix_translation/src/modules/qradar/aql_query_constructor.py
@@ -191,8 +191,8 @@ class AqlQueryStringPatternTranslator:
         operator = self.comparator_lookup[expression.operator]
         expression_01 = self._parse_expression(expression.expr1)
         expression_02 = self._parse_expression(expression.expr2)
-        expression_01_nomap = re.search("^NOMAP", expression_01)
-        expression_02_nomap = re.search("^NOMAP", expression_02)
+        expression_01_nomap = expression_01.startswith("NOMAP")
+        expression_02_nomap = expression_02.startswith("NOMAP")
         if operator == 'AND':
             if expression_01_nomap or expression_02_nomap:
                 self.observation_count -= 1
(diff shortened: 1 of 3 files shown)


stix_shifter/stix_translation/src/modules/qradar/aql_query_constructor.py
```diff
@@ -146,77 +147,142 @@ def _parse_mapped_fields(self, expression, value, comparator, stix_field, mapped
...
+            return ''
 
-            if expression.comparator == ComparisonComparators.NotEqual:
-                comparison_string = self._negate_comparison(comparison_string)
+    @staticmethod
+    def _parse_combined_comparison_expression(self, expression, qualifier=None):
+        operator = self.comparator_lookup[expression.operator]
+        expression_01 = self._parse_expression(expression.expr1)
+        expression_02 = self._parse_expression(expression.expr2)
+        expression_01_nomap = re.search("^NOMAP", expression_01)
```
yurii-klymenko: startswith will work more efficient for this case and 3 other cases

PR #132, opened 2019-04-23
Carbonblack add timestamp and number_observed fields

Collin-Richards:
Added created, modified, first_observed, and last_observed, number_observed fields to carbonblack results. Note some of the tests had to be modified to correctly call the CarbonBlack adapter instead of directly calling the json_to_stix adapter.

Please check the implementation of number_observed is as intended. There is a total_results field in the carbonblack api, but this corresponds to the total number of results that can be returned by the query and not the number of results returned in this instance. Also due to how the json_to_stix_translator.convert_to_stix function works the number_observed should be one for each of the objects in the objects array.


Collin-Richards:
I modified it to throw an exception instead of using an assertion if the number of results is incorrect. I tried adding a test case that would exercise that code, but I failed to come up with one. I believe as long as there's not a bug in the json to stix translator that code shouldn't be hit anyway.

benjamin-craig:
The unit tests failed but I didn't take a look at them yet.

PR #137, opened 2019-05-14
remove nodes of unsupported attributes if ORed

emishas:


mdazam1942:
@delliott90 there are some merge conflicts in couple of test files.

delliott90:
Yeah, I'll handle the conflicts once everything is done on this. It's not quite ready yet to review. The code changes are going to look a little worse than it really is since I renamed a few files to make them consistent with the dummy translation template.

stix_shifter/stix_translation/src/unmapped_attribute_stripper.py
```diff
@@ -0,0 +1,76 @@
+from .patterns.pattern_objects import StartStopQualifier, ObservationExpression, \
+    ComparisonExpression, Pattern, CombinedComparisonExpression, CombinedObservationExpression
+from stix_shifter.stix_translation.src.exceptions import DataMappingException
+
+
+class UnmappedAttributeStripper:
```
mdazam1942: @delliott90 would that be possible to put this `unmapped_attribute_stripper.py` in the /utils folder? or we can create another utils inside the '/stix_translation/src/' ? just for cleanliness cause we have so many files in stix shifter.

delliott90: Yeah, I thought about putting it there but wasn't sure since it's translation-specific and thought it should go under the STIX_translation folder. Though STIX_translation could have its own utils folder and that could contain exceptions, STIX_pattern_parser, transformers, and the unmapped_attribute_stripper.

stix_shifter/stix_translation/src/modules/car/car_data_mapping.py
```diff
@@ -1,110 +1,31 @@
...
 
     def __init__(self, options={}):
-        # use user defined mapping object if provided in options
-        if 'mapping' in options:
-            self.MAPPINGS = options['mapping']
+        self.options = options
+        mapping_json = options.get('mapping')
+        self.MAPPINGS = mapping_json or self.fetch_mapping()
+
+    def fetch_mapping(self):
```
emishas: is * fetch_mapping()* used outside of this class?
you have many copies of this code..
maybe it could go to * BaseDataMapper* as a default implementation and filename parameter?

delliott90: Yeah, I guess if any of the mappers needed a different implementation of it they would just override it.

PR #141, opened 2019-06-14
Update bundle adapter for the pipeline

mdazam1942:
1. Updating setup.py for package not on pypi
2. Updated connection and configuration attributes for the transmission module
3. Return object update
4. Error handling update


setup.py
```diff
@@ -135,8 +135,9 @@
     # For an analysis of "install_requires" vs pip's requirements files see:
     # https://packaging.python.org/en/latest/requirements.html
     install_requires=['stix2-patterns==1.1.0', 'stix2-validator==0.5.0',
-                      'antlr4-python3-runtime==4.7', 'python-dateutil==2.7.3'],  # Optional
+                      'antlr4-python3-runtime==4.7', 'python-dateutil==2.7.3', 'stix2-matcher'],  # Optional
 
+    dependency_links=['git+git://github.com/oasis-open/cti-pattern-matcher.git@master#egg=stix2-matcher'],
```
delliott90: According to https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py
you may need to add the version number. So `...#egg=cti-pattern-matcher-v0.1.0`.
I think the name after egg has to be the repo name instead of the custom name it is now.

mdazam1942: good point. i was also kinda confused about that. i think also have to specify this `cti-pattern-matcher==0.1.0` in install_requires

mdazam1942: But there's also a confusion, the package name is `stix2-matcher` which needs to be set in `install_requires` and whatever package name we provide in there has to be the same as what we set for `#egg=`

PR #143, opened 2019-06-25
Fix ECS adapter integration with ISC

delliott90:
There were some missing method implementations that are used by ISC even though they don't actually do anything when the data source is synchronous. Updated the Elastic Search ECS and synchronous dummy transmission modules.

### DIFF
--- a/stix_shifter/stix_transmission/src/modules/async_dummy/api_client.py
+++ b/stix_shifter/stix_transmission/src/modules/async_dummy/api_client.py
@@ -4,6 +4,7 @@ from ..utils.RestApiClient import RestApiClient
 class APIClient():
 
     def __init__(self, connection, configuration):
+         # Uncomment when implementing data source API client.
         # auth = configuration.get('auth')
         # headers = dict()
         # headers['X-Auth-Token'] = auth.get('token')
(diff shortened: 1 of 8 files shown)


stix_shifter/stix_transmission/src/modules/async_dummy/api_client.py
```diff
@@ -4,6 +4,15 @@
 class APIClient():
 
     def __init__(self, connection, configuration):
+        # auth = configuration.get('auth')
```
mdazam1942: can we put a comment here to specify the reason of commenting out this block of code to avoid confusion of "some" people? it would have been ok if it is a private repo.

stix_shifter/stix_transmission/src/modules/synchronous_dummy/api_client.py
```diff
@@ -0,0 +1,27 @@
+from ..utils.RestApiClient import RestApiClient
+
+
+class APIClient():
+
+    def __init__(self, connection, configuration):
+        # auth = configuration.get('auth')
```
mdazam1942: we should put comment the reason of commenting out this block of code. i know its obvious but since its a public repo some people may miss obvious things :)

PR #149, opened 2019-08-16
Proxy adapter fix

delliott90:
Support for running a proxy host. It can be run from the CLI by navigating to the stix-shifter project on your local dev environment and running:
`python main.py host '{"type": "identity","id": "identity--f431f809-377b-45e0-aa1c-6a4751cae5ff","name": "Proxy Adapter Test", "identity_class": "events"}' "<HOST>:<PORT>"`


main.py
```diff
@@ -159,74 +156,55 @@ def __main__():
...
         @app.route('/translate_results', methods=['POST'])
         def translate_results():
-            request_args = request.get_json( force=True )
-            translation = stix_translation.StixTranslation()
-            print( json.dumps(request_args["results"]))
-            dsl = translation.translate(args.translation_module, 'results', args.data_source, request_args["results"])
-            return json.dumps(dsl)
+            data_source_identity_object = args.data_source
+            host = ProxyHost()
+            return ProxyHost.translate_results(host, data_source_identity_object)
```
mdazam1942: yurri already pointed out about this syntax. can we do this- `return host.translate_results(host, data_source_identity_object)`  as the `host` object of `ProxyHost()` class has been created just before that. And also i think we don't need to pass `host` again as an argument. so the final statement should be-

`return host.translate_results(data_source_identity_object)`

stix_shifter/stix_translation/src/modules/proxy/proxy_translator.py
```diff
@@ -1,20 +1,46 @@
...
-        response = requests.post("http://" + options["proxy_host"] + ":" + options["proxy_port"] + "/translate_results",
-                                 data=json.dumps({"results": data}))
-        return response.text
+    def _unwrap_connection_options(self, connection):
+        connection_options = connection.get('options', {})
+        if connection_options:
+            proxy_auth = connection_options.get('proxy_auth')
+            embedded_connection_options = connection_options.get('options', {})
+            if proxy_auth and embedded_connection_options and embedded_connection_options.get('host'):
+                connection['proxy_auth'] = connection['options'].pop('proxy_auth')
```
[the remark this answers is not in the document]
delliott90: Well, any authentication information could come in as an object couldn't it? We aren't actually using this for anything yet. Do you have any suggestions?

main.py
```diff
@@ -159,74 +156,56 @@ def __main__():
...
+
         @app.route('/transform_query', methods=['POST'])
         def transform_query():
-            request_args = request.get_json( force=True )
-            translation = stix_translation.StixTranslation()
-            dsl = translation.translate(args.translation_module, 'query', args.data_source, request_args["query"])
-            return json.dumps(dsl)
+            print("transforming query")
+            host = ProxyHost()
+            return host.transform_query(host)
```
mdazam1942: do we still need `host` object to pass in as argument?

delliott90: No, looks like I missed that one.

PR #162, opened 2019-09-05
Big fix v2.0

Raghuvaran-krishnan:
BigFix Version 2.0

==========================================
Features Implemented
==========================================
Add support for Combined observation expression.
Added Qualifier support in the query for example- START/STOP, any time range.
Added wildcards supports.
Verified if all applicable observation operator and comparison operator supports are implemented
Supported operators:
1. = 
2. !=
3. LIKE
4. MATCHES
5. ">"
6. ">="
7. "<="
8. "<"
9. IN
File object and its associated atrributes supported on Data Source:
	Name
	Hashes(sha256|md5|sha1)
	Folder
	size
Process object and its associated attributes supported on Data Source:
	Name
	process id(pid)
	parent process id(ppid)
	User of process ==> Reference to User object
	Binary ref ==> Reference to File object
	
Socket object and its associated attributes supported on Data Source:
	Local address
	Local port
	Remote address
	Remote port
	
Network Adapter object and its associated atrributes supported on Data Source
	mac address ==> refers to local mac address


stix_shifter/stix_translation/src/modules/bigfix/README.md
```diff
@@ -1,180 +1,337 @@
...
+  2. `("Address", address of it as string | "n/a",  mac address of it as string | "n/a") of adapters whose ((mac address of it as string = "0a-65-a4-7f-ad-88" as string) AND (loopback of it = false AND address of it != "0.0.0.0")) of network`
+
+### Translate, Transmit, Translate Result flow of a STIX pattern:
 ```
-[process:name = 'system']
+[ipv4-addr:value LIKE = '192']
```
delliott90: I think you meant to have either `LIKE` or `=` rather than both.

stix_shifter/stix_translation/src/modules/bigfix/json/from_stix_map.json
```diff
@@ -0,0 +1,49 @@
+{
+  "file": {
+    "fields": {
+      "name": ["file.name"],
+      "hashes.SHA-256": ["file.sha256"],
+      "hashes.SHA-1": ["file.sha1"],
+      "hashes.md5": ["file.md5"],
```
delliott90: The STIX md5 attribute should be uppercase here. Previous hash mappings in STIX-shifter have followed the example in the STIX documentation as seen in https://docs.google.com/document/d/1nK1RXcE2aMvQoG1Kgr3aTBtHZ1IyehzOk7vU0n5FUGY/edit#heading=h.jphmvccraytb.

stix_shifter/stix_translation/src/modules/bigfix/json/from_stix_map.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "process": {
+    "fields": {
+      "name": ["process.name"],
+      "pid": ["process.pid", "process.process id"],
+      "parent_ref.pid": ["process.ppid"],
+      "binary_ref.name" : ["file.name"],
+      "binary_ref.hashes.SHA-256" : ["file.sha256"],
+      "binary_ref.hashes.SHA-1": ["file.sha1"],
+      "binary_ref.hashes.md5": ["file.md5"],
```
delliott90: Uppercase `md5`

stix_shifter/stix_translation/src/modules/bigfix/README.md
```diff
@@ -1,180 +1,337 @@
...
 
 ```
-[{"computer_identity": "12369754-bigdata4545.canlab.ibm.com", "subQueryID": 1, "start_time": "1541424881", "type": "process", "process_name": "systemd", "process_id": "1", "sha256hash": "74c4ff75e3623e64e3d6620864b69ed1d75fa460e520b88ed234234fsdfsdsdfs", "sha1hash": "916933045c5c91ebcaa325e7f8302f3123123dfgf0000", "md5hash": "28a9beb86c4d4c31ba572805baaa777f", "file_path": "/file/path/systemd"}]
+### A search_id is returned which is further passed to STIX transmission result module 
+```
+{'search_id': '4191', 'success': True}
+```
+### Transmit get result
 ```
+transmit bigfix "{\"host\":\"xx.xx.xx.xx\", \"port\":\"xxxxx\", \"cert_verify\":\"False\"}" {\"auth\":{\"username\":\"hclapi\",\"password\":\"jnGfdEYY8T$e\"}} results 4191 <offset> <length>
```
mdazam1942: please sanitize the username and password. put something random.

stix_shifter/stix_translation/src/modules/bigfix/README.md
```diff
@@ -1,180 +1,337 @@
...
+
+```
+### A search_id is returned which is further passed to STIX transmission result module 
+```
+{'search_id': '4192', 'success': True}
+```
+### Transmit get result
+```
+transmit bigfix "{\"host\":\"xx.xx.xx.xx\", \"port\":\"xxxxx\", \"cert_verify\":\"False\"}" {\"auth\":{\"username\":\"hclapi\",\"password\":\"jnGfdEYY8T$e\"}} results 4192 <offset> <length>
```
mdazam1942: please sanitize the username and password. put something random.

stix_shifter/stix_translation/src/modules/bigfix/README.md
```diff
@@ -1,180 +1,337 @@
...
+"<BESAPI xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:noNamespaceSchemaLocation=\"BESAPI.xsd\"><ClientQuery><ApplicabilityRelevance>true</ApplicabilityRelevance><QueryText>(\"file\", name of it | \"n/a\",  \"sha256\", sha256 of it | \"n/a\",  \"sha1\", sha1 of it | \"n/a\",  \"md5\", md5 of it | \"n/a\",  pathname of it | \"n/a\",  size of it | 0,  (modification time of it - \"01 Jan 1970 00:00:00 +0000\" as time)/second) of files whose ((modification time of it is greater than or equal to \"10 Jan 2013 08:43:10 +0000\" as time AND modification time of it is less than or equal to \"23 Oct 2019 10:43:10 +0000\" as time)) of folder \"/tmp\"</QueryText><Target><CustomRelevance>true</CustomRelevance></Target></ClientQuery></BESAPI>"
+
+```
+### A search_id is returned which is further passed to STIX transmission result module 
+```
+{'search_id': '4193', 'success': True}
+```
+### Transmit get result
+```
+transmit bigfix "{\"host\":\"xx.xx.xx.xx\", \"port\":\"xxxxx\", \"cert_verify\":\"False\"}" {\"auth\":{\"username\":\"hclapi\",\"password\":\"jnGfdEYY8T$e\"}} results 4193 <offset> <length>
```
mdazam1942: please sanitize the username and password. put something random.

stix_shifter/stix_translation/src/modules/bigfix/json/to_stix_map.json
```diff
@@ -73,9 +102,70 @@
             "cybox": false
         }
     ],
-    "computer_identity":
-    {
+    "computer_identity": {
```
mdazam1942: I know this has been there before. we need to set `computer_identity` field inside a custom stix object instead of setting as name object. bigfix custom stix object can be named as `x_com_bigfix_relevance`. there are some examples of how to add a custom object in QRadar and Splunk  to_stix_map.json

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -64,26 +67,27 @@ def create_status_connection(self, search_id):
                         if total_results <= 0:
                             return_obj['status'] = Status.ERROR.value
 
-                    if return_obj['success'] == True and return_obj['status'] == Status.COMPLETED.value:
+                    if return_obj['success'] and return_obj['status'] == Status.COMPLETED.value:
```
mdazam1942: do you have any query that sets the status COMPLETE? I have tried many queries all of them stayed RUNNING forever even though we have got results back already. UDS will automatically cancel the query when status stuck at RUNNING state. We need a bail out condition here to set the status to COMPLETE if we have received all the results that are available.  we can discuss further details if you need.

stix_shifter/stix_translation/src/modules/bigfix/json/to_stix_map.json
```diff
@@ -31,14 +31,43 @@
...
+      },
+      {
+        "key": "process.creator_user_ref",
+        "object": "process",
+        "references": "user"
+      }
+    ],
     "start_time": [
         {
             "key": "created",
```
mdazam1942: we are removing the `created` field from the mapping file and setting the `created` and `modified` timestamp as the current time. we are merging an update shortly in master branch. please pull it down the latest change when it is merged.

stix_shifter/stix_translation/src/modules/bigfix/README.md
Raghuvaran-krishnan: Incorporated with latest Commit.

stix_shifter/stix_translation/src/modules/bigfix/README.md
Raghuvaran-krishnan: Incorporated with latest Commit.

stix_shifter/stix_translation/src/modules/bigfix/json/from_stix_map.json
Raghuvaran-krishnan: Incorporated with latest Commit.

stix_shifter/stix_translation/src/modules/bigfix/json/from_stix_map.json
Raghuvaran-krishnan: Incorporated with latest Commit.

stix_shifter/stix_translation/src/modules/bigfix/README.md
Raghuvaran-krishnan: Incorporated with latest Commit.

mdazam1942:
@Raghuvaran-krishnan i'm having hard time testing transmission query. Same stix pattern `python main.py translate bigfix query '{}' "[process:name LIKE 'dnsmasq']"` gave me results back previously but not getting anything back after your latest changes. while debugging i found it's responding with one error `"The operator \"start time\" is not defined."` i have seen this error before. is there any way to avoid this error being hit? are you setting anything in the default query format that triggers this error?

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: Incorporated the changes that Danny had suggested in his mail.
Now internally the status query will check the status for a period of  (timeout - 6 seconds) (if timeout = 1min, will check for 54 sec) sec and if progress doesn't reach 100% will return back a status of **'Completed'** and along with the last observed progress percentage.

Raghuvaran-krishnan:
> 
> 
> @Raghuvaran-krishnan i'm having hard time testing transmission query. Same stix pattern `python main.py translate bigfix query '{}' "[process:name LIKE 'dnsmasq']"` gave me results back previously but not getting anything back after your latest changes. while debugging i found it's responding with one error `"The operator \"start time\" is not defined."` i have seen this error before. is there any way to avoid this error being hit? are you setting anything in the default query format that triggers this error?
==============================================================
The behavior being experienced is as expected and not because of recent changes
Please find below the explanation for the same:

- The STIX pattern shared doesn't specify a start stop Qualifier so by default a start time = current time and stop time = current time -5 minutes is appended to BigFix relevance query.
- The start time parameter for the currently running "dnsmasq" processes doesnot fall under the default time range window of BigFix relevance query
- Hence results = []

- Howerever If given a stix pattern of 

**translate
bigfix
query
‘{}’
"[process:name LIKE 'dnsmasq'] START t'2019-06-01T07:03:17Z' STOP t'2019-09-13T07:03:17Z'"**

- Start time of currently running "dnsmasq" processes are within the above Qualifier range. This returns results

_{'success': True, 'data': [
{'computer_identity': '537366757-farpoint.hcl.local', 'subQueryID': 1, 'sha256hash': '31e96c3cf483177865830298305e55cbd8bf7afebecc6bcba78360133cf24140', 'sha1hash': '2043de0a76149d0b9e5a0ee0183c077c9235c1c8', 'md5hash': '05546846517405bcc46c4176c4ddf03a', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '1', 'process_user': 'nobody', 'timestamp': '1559567777', 'process_name': 'dnsmasq', 'process_id': '3170', 'file_size': '344856', 'type': 'process', 'event_count': '1'},
 {'computer_identity': '537366757-farpoint.hcl.local', 'subQueryID': 1, 'sha256hash': '31e96c3cf483177865830298305e55cbd8bf7afebecc6bcba78360133cf24140', 'sha1hash': '2043de0a76149d0b9e5a0ee0183c077c9235c1c8', 'md5hash': '05546846517405bcc46c4176c4ddf03a', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '3170', 'process_user': 'root', 'timestamp': '1559567777', 'process_name': 'dnsmasq', 'process_id': '3174', 'file_size': '344856', 'type': 'process', 'event_count': '1'},
 {'computer_identity': '3665364-quark.hcl.local', 'subQueryID': 1, 'sha256hash': 'bb4deebc8e96f0925de80a178327d89d6e982a33350c35ed2299b8c0ceb19080', 'sha1hash': '591a1b9686350564c8e34e32486ebf3e323ad032', 'md5hash': 'cdbf7b857fb22b045cc2674b037b9b48', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '1', 'process_user': 'nobody', 'timestamp': '1565695303', 'process_name': 'dnsmasq', 'process_id': '1527', 'file_size': '344928', 'type': 'process', 'event_count': '1'}]}_
 
 Hope this clarifies.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_api_client.py
```diff
@@ -1,8 +1,10 @@
 import base64
 from ..utils.RestApiClient import RestApiClient
 
+TIMEOUT = 60
```
delliott90: The connection timeout is stored on UDS as minutes, not seconds. The default value is 1, but it can always be changed when the data source gets added. If the BigFix API requires seconds for the timeout value, you'll need to do the conversion before creating the client.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -51,39 +49,51 @@ def create_status_connection(self, search_id):
...
+        return_obj = dict()
+        try:
+            response = self.api_client.get_sync_query_results(self.RELEVANCE)
+            response_txt = response.read().decode('utf-8')
+            client_count = self.DEFAULT_CLIENT_COUNT
+            search = re.search(self.PATTERN, response_txt, re.IGNORECASE)
+            if search:
+                client_count = search.group(1)
+            client_count = int(client_count)
+            return_obj = self.status_api_response(search_id, client_count)
```
mdazam1942: if I'm not wrong this looks like its not gonna return anything to the UDS until the timeout limit reaches for all the RUNNING status cases. we need to return the status whenever UDS makes a status call even if its running. UDS itself make calls to check the status until the timeout.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -51,39 +49,51 @@ def create_status_connection(self, search_id):
...
                             return_obj['status'] = Status.ERROR.value
-
-                    if return_obj['success'] == True and return_obj['status'] == Status.COMPLETED.value:
-                        if not self.get_succes_status(response_dict):
-                            return_obj['status'] = Status.ERROR.value
-                            ErrorResponder.fill_error(return_obj, response_dict, ['results', '+isFailure=True','~result'])
-
+                        break
+                    time_iter -= WAIT_PERIOD
+                    if time_iter < WAIT_PERIOD:
```
mdazam1942: isn't it safe to track the timestamp instead of counting wait period?

mdazam1942:
> > @Raghuvaran-krishnan i'm having hard time testing transmission query. Same stix pattern `python main.py translate bigfix query '{}' "[process:name LIKE 'dnsmasq']"` gave me results back previously but not getting anything back after your latest changes. while debugging i found it's responding with one error `"The operator \"start time\" is not defined."` i have seen this error before. is there any way to avoid this error being hit? are you setting anything in the default query format that triggers this error?
> > ==============================================================
> > The behavior being experienced is as expected and not because of recent changes
> > Please find below the explanation for the same:
> 
> * The STIX pattern shared doesn't specify a start stop Qualifier so by default a start time = current time and stop time = current time -5 minutes is appended to BigFix relevance query.
> * The start time parameter for the currently running "dnsmasq" processes doesnot fall under the default time range window of BigFix relevance query
> * Hence results = []
> * Howerever If given a stix pattern of
> 
> **translate bigfix query ‘{}’ "[process:name LIKE 'dnsmasq'] START t'2019-06-01T07:03:17Z' STOP t'2019-09-13T07:03:17Z'"**
> 
> * Start time of currently running "dnsmasq" processes are within the above Qualifier range. This returns results
> 
> _{'success': True, 'data': [ {'computer_identity': '537366757-farpoint.hcl.local', 'subQueryID': 1, 'sha256hash': '31e96c3cf483177865830298305e55cbd8bf7afebecc6bcba78360133cf24140', 'sha1hash': '2043de0a76149d0b9e5a0ee0183c077c9235c1c8', 'md5hash': '05546846517405bcc46c4176c4ddf03a', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '1', 'process_user': 'nobody', 'timestamp': '1559567777', 'process_name': 'dnsmasq', 'process_id': '3170', 'file_size': '344856', 'type': 'process', 'event_count': '1'}, {'computer_identity': '537366757-farpoint.hcl.local', 'subQueryID': 1, 'sha256hash': '31e96c3cf483177865830298305e55cbd8bf7afebecc6bcba78360133cf24140', 'sha1hash': '2043de0a76149d0b9e5a0ee0183c077c9235c1c8', 'md5hash': '05546846517405bcc46c4176c4ddf03a', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '3170', 'process_user': 'root', 'timestamp': '1559567777', 'process_name': 'dnsmasq', 'process_id': '3174', 'file_size': '344856', 'type': 'process', 'event_count': '1'}, {'computer_identity': '3665364-quark.hcl.local', 'subQueryID': 1, 'sha256hash': 'bb4deebc8e96f0925de80a178327d89d6e982a33350c35ed2299b8c0ceb19080', 'sha1hash': '591a1b9686350564c8e34e32486ebf3e323ad032', 'md5hash': 'cdbf7b857fb22b045cc2674b037b9b48', 'file_path': '/usr/sbin/dnsmasq', 'process_ppid': '1', 'process_user': 'nobody', 'timestamp': '1565695303', 'process_name': 'dnsmasq', 'process_id': '1527', 'file_size': '344928', 'type': 'process', 'event_count': '1'}]}_
> 
> Hope this clarifies.

@Raghuvaran-krishnan i understand the result can return empty []. I tried both with or without the start/stop qualifier. But my question is what would be the reason to get this error `"The operator \"start time\" is not defined."` if query can't retrieve any data? Am I passing any invalid syntax in the query? here's the query-

Raghuvaran-krishnan:
> 
> 
> > > The behavior being experienced is as expected and not because of recent changes
> > > Please find below the explanation for the same:
> > 
> > 
> > 
> > * The STIX pattern shared doesn't specify a start stop Qualifier so by default a start time = current time and stop time = current time -5 minutes is appended to BigFix relevance query.
> > * The start time parameter for the currently running "dnsmasq" processes doesnot fall under the default time range window of BigFix relevance query
> > * Hence results = []
> > * Howerever If given a stix pattern of
> > 
> > **translate bigfix query ‘{}’ "[process:name LIKE 'dnsmasq'] START t'2019-06-01T07:03:17Z' STOP t'2019-09-13T07:03:17Z'"**
> > 
> > * Start time of currently running "dnsmasq" processes are within the above Qualifier range.
> >   Hope this clarifies.
> 
> @Raghuvaran-krishnan i understand the result can return empty [] if query can't find any running process during that particular time. I tried both with or without the start/stop qualifier.
> `python main.py translate bigfix query '{}' "[process:name = 'dnsmasq'] START t'2019-09-20T19:14:00.000Z' STOP t'2019-09-20T19:19:00.000Z'"`
> 
> But my question is. if query can't retrieve any data than what would be the reason to get this error `"The operator \"start time\" is not defined."` ? Hope I am not passing any invalid syntax cause i just copied the translated query from CLI. here's the query-
> 
> `python main.py transmit bigfix '{"cert_verify":"False","host": "3.214.172.161", "port":52311}' '{"auth": {"username": "xxxxx","password": "xxxxxx"}}' query '<BESAPI xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="BESAPI.xsd"><ClientQuery><ApplicabilityRelevance>true</ApplicabilityRelevance><QueryText>("process", name of it | "n/a", pid of it as string | "n/a", "sha256", sha256 of image file of it | "n/a", "sha1", sha1 of image file of it | "n/a", "md5", md5 of image file of it | "n/a", pathname of image file of it | "n/a", ppid of it as string | "n/a", (if (windows of operating system) then user of it as string | "n/a" else name of user of it as string | "n/a"), size of image file of it | 0, (if (windows of operating system) then (creation time of it | "01 Jan 1970 00:00:00 +0000" as time - "01 Jan 1970 00:00:00 +0000" as time)/second else (start time of it | "01 Jan 1970 00:00:00 +0000" as time - "01 Jan 1970 00:00:00 +0000" as time)/second)) of processes whose (((name of it as lowercase = "dnsmasq" as lowercase)) AND (if (windows of operating system) then (creation time of it | "01 Jan 1970 00:00:00 +0000" as time is greater than or equal to "20 Sep 2019 19:14:00 +0000" as time AND creation time of it | "01 Jan 1970 00:00:00 +0000" as time is less than or equal to "20 Sep 2019 19:19:00 +0000" as time) else (start time of it | "01 Jan 1970 00:00:00 +0000" as time is greater than or equal to "20 Sep 2019 19:14:00 +0000" as time AND start time of it | "01 Jan 1970 00:00:00 +0000" as time is less than or equal to "20 Sep 2019 19:19:00 +0000" as time)))</QueryText><Target><CustomRelevance>true</CustomRelevance></Target></ClientQuery></BESAPI>'`

Hi @mdazam1942,
      This issue is noticed on BigFix01 and ds9.hcl.local nodes connected to the BigFix server in test environment.
This is not pertaining to any issue with Query. It may be due to some other issue with this client.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: @mdazam1942 
Can you please clarify provide more details on this query?

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: Hi Azam,

- The DS(data source) API does not provide a timeout parameter with the API calls.  The DS(data source) does not provide status parameter in Asynchronous API calls. There is no persistence of data in connector.
- Hence had implemented an internal wait logic in the status call in the form of a while loop which will accumulate data from data source and check progress against internal timeout(1 minute) with wait time of 6 seconds per loop (or seconds to timeout) whichever is smaller.
- This will immediately send status back if progress reaches 100% before timeout is reached.
- If progress failed to reach 100% it will change the status to completed and send the status back with progress equal to last observed progress.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_api_client.py
Raghuvaran-krishnan: Hi Danny,
                The BigFix API does not support/provide way to pass timeout parameter.
                We have implemented an internal wait logic in the status call to BigFix API.
                We could convert it to seconds

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
mdazam1942: We are clear about the API limitations on timeout. The main concern is, we can't keep the UDS caller waiting to send back the status if progress is less than 100%. That may result in cloud functions timeout. We need to find a way to persist the internal wait logic/count and return the status back to UDS at the same time. That's why I suggested to use the timestamp instead in my other comment.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -4,12 +4,17 @@
 import json
 import re
 import sys
+import time
 from .....utils.error_response import ErrorResponder
 import xmltodict
 
+WAIT_PERIOD = 6
```
mdazam1942: just wondering why `WAIT_PERIOD` is  `6` . is it just an arbitrary number?

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: As connectors are invoked via cloud functions we may not be able to persist data on the connector side.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: Yes it is an arbitrary value used for the wait logic.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
mdazam1942: Right. You wont be able to persist the timeout in connector side. But, Its the connector's responsibility to return the status back to UDS on time. This is very specific case for BigFix so it needs to be handled in connector side.

delliott90:
I'm currently working on a defect for QRadar where a pattern with `NOT` in it wasn't getting properly negated in the AQL query. An example of such a pattern would be `[url:value NOT = 'www.example.ca']`. Can you confirm this is supported in the BigFix adapter?

Raghuvaran-krishnan:
> 
> 
> I'm currently working on a defect for QRadar where a pattern with `NOT` in it wasn't getting properly negated in the AQL query. An example of such a pattern would be `[url:value NOT = 'www.example.ca']`. Can you confirm this is supported in the BigFix adapter?

HI @delliott90 ,
                       We have implemented the code for negation operator in BigFix with our latest commit.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: As discussed and shared over email we have incorporated the query status completion logic with the latest commit.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
Raghuvaran-krishnan: HI Azam,
As discussed and shared over email we have incorporated the query status completion logic with the latest commit.

mdazam1942: looks good. thanks!

PR #167, opened 2019-09-10
remove CIM from-STIX mappings that were mapped to none

delliott90:
This was failing for some Splunk pattern translations due to pattern mappings being attributed to none. Example pattern that was failing:
`python main.py translate splunk query '{}' "[artifact:payload_bin = 'tealdkfjasdflasdj']"`

### DIFF
--- a/stix_shifter/stix_translation/src/modules/cim/json/from_stix_map.json
+++ b/stix_shifter/stix_translation/src/modules/cim/json/from_stix_map.json
@@ -1,6 +1,4 @@
 {
-    "artifact": "None",
-    "as": "None",
     "directory": {
         "cim_type": "endpoint",
         "fields": {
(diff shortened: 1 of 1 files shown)


jleemur:
Looks good, thanks

ph34tur3:
Hi @delliott90 ,
I have got a question regarding a specific line in _stix_shifter/stix_translation/src/modules/cim/json/from_stix_map.json_.

In line 92 (`"dst_ref.value": "src",`) the translation maps from **destination** to **source** which bugs me since line 90 (`"src_ref.value": "src",`) already maps from **source** to **source** and line 93 (`"dst_port": "dest_port",`) maps to a **destination** port.

Is this translation correct?
Thanks in advance.

delliott90:
@ph34tur3 I would say that's a defect in the mapping. It should probably be mapping to `dst`.

PR #186, opened 2019-10-17
MS Windows Defender ATP

aarthips:
Implemented UDS connector for MS Windows Defender ATP.


requirements-dev.txt
```diff
@@ -18,4 +18,5 @@ pytest-cov==2.5.1
 six==1.11.0
 wrapt==1.10.11
 requests==2.20.1
-flask==1.0.2
\ No newline at end of file
+flask==1.0.2
+adal==1.2.2
```
mdazam1942: looks like this `adal==1.2.2` is an external library only used by msatp. you should put this in requirements.txt instead and also in setup.py.

delliott90:
I didn't see it in the from_stix_map, could you add mappings for `x-com-msatp:computer_identity` and `directory:path`?

tests/stix_translation/test_msatp_stix_to_query.py
```diff
@@ -0,0 +1,218 @@
+from stix_shifter.stix_translation import stix_translation
+import unittest
+import re
+
+translation = stix_translation.StixTranslation()
+
+
+def _remove_timestamp_from_query(queries):
```
delliott90: What's the purpose of removing the timestamps for each test? Is it because the current time keeps changing. I guess I can see this when testing the default time ranges, but why is it needed when you specify START STOP qualifiers in the test pattern?

stix_shifter/stix_translation/src/modules/msatp/query_constructor.py
```diff
@@ -0,0 +1,406 @@
...
+        :return: str
+        """
+        if expression_operator not in self.comparator_lookup:
+            raise NotImplementedError(
+                "Comparison operator {} unsupported for MSATP adapter".format(expression_operator.name))
+        return self.comparator_lookup.get(expression_operator)
+
+    def _parse_expression(self, expression, qualifier=None):
+        """
+        Complete formation of relevance query from ANTLR expression object
```
mdazam1942: I see the word `relevance` is used in many places of query_constructor.py file. I think it is copied from BigFix(relevance is the query language of BigFix). you need to replace that with whatever is appropriate for msatp. Usually we use the query name of the api or something that refers the datasource/api.

mdazam1942:
@aarthips we pushed some changes in our master branch to fix our generic restapi client. please rebase the latest commit in your branch and make necessary api client change. you probably have to just change the api_client.py initialization.

stix_shifter/stix_translation/src/modules/msatp/query_constructor.py
```diff
@@ -0,0 +1,406 @@
...
+    :param options: dict,timerange defaults to 5
+    :return: str, KQL query
+    """
+    timerange = options['timerange']
+    final_queries = []
+    translated_dictionary = QueryStringPatternTranslator(pattern, data_model_mapper, timerange)
+    final_query = translated_dictionary.qualified_queries
+    if len(final_query) > 1:
+        final_queries.append('union {}'.format(','.join(final_query)))
+        return final_queries
```
delliott90: You could return the `final_queries` list with just one query string.

stix_shifter/stix_translation/src/modules/msatp/stix_to_query.py
```diff
@@ -0,0 +1,30 @@
...
+        KQL. This defaults to the from_stix_map.json in the stix_shifter/stix_translation/src/modules/msatp/json/
+        directory
+        :type mapping: str (filepath)
+        :return: kql query string
+        :rtype: str
+        """
+
+        logger.info("Converting STIX2 Pattern to KQL query")
+
+        # TODO: Will need to implement a data_model_mapper before using the unmapped attribute stripper
```
delliott90: This TODO can be removed, I think it was in the Bigfix code from an earlier fix.

stix_shifter/stix_transmission/src/modules/msatp/msatp_connector.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def ping(self):
+        """Ping the endpoint."""
+        response_txt = None
+        return_obj = dict()
+        try:
+            if self.init_error is not None:
+                raise Exception(self.init_error)
+            response = self.api_client.ping_box()
+            return self._handle_errors(response, return_obj)
```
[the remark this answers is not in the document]
subash0993: Updated ping implementation to return status and formatted error in case of failure.

requirements-dev.txt
subash0993: I have removed adal library for this file and added to requirements.txt and setup.py

tests/stix_translation/test_msatp_stix_to_query.py
subash0993: This is implemented for default timestamp

stix_shifter/stix_translation/src/modules/msatp/query_constructor.py
subash0993: Updated relevance in query_constructor.py

stix_shifter/stix_translation/src/modules/msatp/stix_to_query.py
subash0993: TODO has been removed in stix_to_query.py

stix_shifter/stix_translation/src/modules/msatp/query_constructor.py
subash0993: Code updated

subash0993:
> I didn't see it in the from_stix_map, could you add mappings for x-com-msatp:computer_identity and directory:path?

Hi @delliott90 Code Updated to handle the custom attribute x-com-msatp and directory:path. Stix attributes "directory:path" and "parent_directory_ref.path" are mapped to "FolderPath" or "InitiatingProcessFolderPath" attribute in MSATP data source. These datasource attributes display the path along with the file name. As per the stix documentation path attribute will not take file name in the STIX pattern. Hence, the operators =, !=, MATCHES, IN are not supported in STIX pattern for 'path' attribute, only the LIKE operator is supported. We handled to throw error message for other operators.

subash0993:
> @aarthips we pushed some changes in our master branch to fix our generic restapi client. please rebase the latest commit in your branch and make necessary api client change. you probably have to just change the api_client.py initialization.

Hi @mdazam1942 We rebased the code for api client change.

PR #191, opened 2019-10-22
wip: sni certificate check refactor

benjamin-craig:


stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
```diff
@@ -51,8 +53,21 @@ def call_api(self, endpoint, method, headers=None, params=[], data=None, urldata
             else:
                 url = 'https://' + self.server_ip + '/' + endpoint
             try:
-                call = getattr(requests, method.lower())
-                response = call(url, headers=actual_headers, cert=self.cert_file_name, data=data, verify=self.cert_verify)
+                s = requests.Session()
```
mdazam1942: lets call it `req_session` or `session` instead of just `s`

PR #193, opened 2019-10-24
Fix certs

benjamin-craig:


stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
```diff
@@ -6,31 +6,61 @@
...
+            if self.server_cert_content is not None:
+                with open(self.server_cert, 'w') as f:
                     try:
-                        f.write(self.cert)
+                        f.write(self.server_cert_content)
                     except IOError:
                         print('Failed to setup certificate')
+                self.server_cert_content = self.server_cert
+            else:
+                self.server_cert_content = None
```
delliott90: Isn't `self.server_cert_content` already `None`?

stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
```diff
@@ -47,24 +77,34 @@ def call_api(self, endpoint, method, headers=None, params=[], data=None, urldata
...
+                        and "Deprecated" in response.headers['Content-Type']:
+                    print("WARNING: " +
+                          response.headers['Content-Type'], file=sys.stderr)
                 return ResponseWrapper(response)
             except Exception as e:
                 print('exception occured during requesting url: ' + str(e))
                 raise e
         finally:
-            if self.cert_file_name is not None:
+            if self.server_cert_exist:
```
delliott90: Can't you just check `self.server_cert_content` rather than creating this extra variable?

stix_shifter/stix_transmission/src/modules/qradar/arielapiclient.py
```diff
@@ -53,7 +55,7 @@ def __init__(self, connection, configuration):
                                     connection.get('cert', None),
                                     headers,
                                     url_modifier_function,
-                                    connection.get('cert_verify', 'True')
+                                    connection.get('selfSignedCert', True)
```
delliott90: This change would need to happen for all adapters, correct? It's missing from the dummy modules, elastic_ecs, and Splunk.

benjamin-craig: Yes Azam is going to extend it I hope

stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
benjamin-craig: Yea Azam is going over the logic I wrote it fast just so we could get it  into a PR

stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
benjamin-craig: Yea for sure

PR #196, opened 2019-10-29
SNI support and splunk adaptor calls the rest api client multiple times with single rest api client object

benjamin-craig:


benjamin-craig:
@mdazam1942 I removed the confusing logic to have 2 variables with the same content.

stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
```diff
@@ -23,29 +23,29 @@ def __init__(self, host, port=None, cert=None, headers={}, url_modifier_function
         self.sni = None
         #Gateway Case -- use client cert cert_verify is None
         if mutual_auth:
-            self.server_cert_content = None
+            self.server_cert_call_value = None
```
mdazam1942: actually `self.server_cert_content` is more meaningful than `self.server_cert_call_value`. you don't need to change this variable name i think

PR #199, opened 2019-11-01
Big fix v2.0

Raghuvaran-krishnan:
Changes
========
1. transformers.py
	1. Modified the existing code to return the parent directory_path using python os module.

2. to_stix_map.json
	1. Added transformer "ToFilePath" to file:directory_path STIX_attribute to remove the filename from the path value.
	2. Removed network-traffic attribute "resolved_to_ref" for mac-addr:value STIX attribute as BigFix doesnot provided network-traffic details while querying for mac-addr of local computer/machine.
	
3. 	bigfix_api_client.py
	1. Created a new instance of RestApiClient as 2 subsequent API calls were throwing SSL error.
	Verified the working of new code.
	
4. bigfix_results_connector.py
	1. Modified the status parameter to return 'true' if the API call retuned a response code of 200 but no data is returned from BigFix server.
	2. Modified the mac address query result method to fix --stix-validator issue
		
5. bigfix_status_connector.py
	1. Removed a code that returned a status = 'ERROR' even if API call returned a response code of 200.

6. test_bigfix.py
       Modified unit test case code for above status query changes.


stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -57,8 +57,6 @@ def update_query_status(self, return_obj, response_dict, client_count):
         return_obj = self._get_progress_status(client_count, reporting_agents, return_obj)
         if client_count <= reporting_agents:
             return_obj['status'] = Status.COMPLETED.value
-            if total_results <= 0:
-                return_obj['status'] = Status.ERROR.value
         else:
```
mdazam1942: We are having some issue in pipeline testing while the progress is less than 100 but the status is COMPLETED. We are getting some conflicting status. I know we have discussed about this a lot earlier. Can put a check before returning the status to set the progress value to 100 when the status is COMPLETED? we can still keep the progress threshold check the same.

Raghuvaran-krishnan:
A new PR has been been raised hence closing this request #202

PR #202, opened 2019-11-05
Updated Bigfix adapter with review comments

aarthips:
1.	transformers.py
i.	Modified the existing code to return the parent directory_path using python os module.
2.	to_stix_map.json
i.	Added transformer "ToFilePath" to file:directory_path STIX_attribute to remove the filename from the path value.
ii.	Removed network-traffic attribute "resolved_to_ref" for mac-addr:value STIX attribute as BigFix doesnot provided network-traffic details while querying for mac-addr of local computer/machine.
3.	bigfix_api_client.py
i.	Created a new instance of RestApiClient as 2 subsequent API calls were throwing SSL error.
Verified the working of new code.
4.	bigfix_results_connector.py
i.	Modified the status parameter to return 'true' if the API call retuned a response code of 200 but no data is returned from BigFix server.
ii.	Modified the mac address query result method to fix --stix-validator issue
iii.	Modified progress percentage to 100% for “Completed” status
5.	bigfix_status_connector.py
i.	Removed a code that returned a status = 'ERROR' even if API call returned a response code of 200.
6.	test_bigfix.py
Modified unit test case code for above status query changes.
7.	test_car_json_to_stix.py
Modified unit test case for directory path value

### DIFF
--- a/stix_shifter/stix_translation/src/utils/transformers.py
+++ b/stix_shifter/stix_translation/src/utils/transformers.py
@@ -3,6 +3,7 @@ from datetime import datetime, timezone, tzinfo, timedelta
 import base64
 import socket
 import re
+import os
 from urllib.parse import urlparse
 
 
@@ -116,7 +117,7 @@ class ToFilePath(ValueTransformer):
     @staticmethod
     def transform(obj):
         try:
-            return obj[0:len(obj) - len(re.split(r'[\\/]', obj)[-1])]
+            return os.path.dirname(obj)
         except ValueError:
             print("Cannot convert input to path string")
 
--- a/stix_shifter/stix_translation/src/utils/transformers.py
+++ b/stix_shifter/stix_translation/src/utils/transformers.py
@@ -117,9 +117,20 @@ class ToFilePath(ValueTransformer):
     @staticmethod
     def transform(obj):
         try:
-            return os.path.dirname(obj)
+            return obj[0:len(obj) - len(re.split(r'[\\/]', obj)[-1])]
         except ValueError:
             print("Cannot convert input to path string")
+
+
+class ToDirectoryPath(ValueTransformer):
+    """A value transformer for expected directory path"""
+
+    @staticmethod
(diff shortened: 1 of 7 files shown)


stix_shifter/stix_translation/src/utils/transformers.py
```diff
@@ -116,7 +117,7 @@ class ToFilePath(ValueTransformer):
     @staticmethod
     def transform(obj):
         try:
-            return obj[0:len(obj) - len(re.split(r'[\\/]', obj)[-1])]
+            return os.path.dirname(obj)
```
mdazam1942: not sure why you changed to `os.path.dirname(obj)`. i dont see you get correct file path and also its breaking some other module path transformation.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_status_connector.py
```diff
@@ -57,13 +57,13 @@ def update_query_status(self, return_obj, response_dict, client_count):
...
             return_obj['status'] = Status.COMPLETED.value
-            if total_results <= 0:
-                return_obj['status'] = Status.ERROR.value
+            return_obj['progress'] = 100
         else:
             if return_obj['progress'] > PROGRESS_THRESHOLD:
                 sleep_time = math.ceil((100 - return_obj['progress']) * 10 / 100)
                 time.sleep(sleep_time)
                 return_obj['status'] = Status.COMPLETED.value
+                return_obj['progress'] = 100
```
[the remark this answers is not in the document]
aarthips: Updated the code to remove sleep_time and total_results

stix_shifter/stix_translation/src/utils/transformers.py
aarthips: This was added to get directory path without trailing "\" or "/" slash. I have updated other impacted tests cases in my earlier commit. I have added new transformer function for bigfix and msatp implementation and reverted the other test changes.

PR #207, opened 2019-11-13
Fix bundle ping response and api call

mdazam1942:
1. Ping didn't return correct status. Making a head call to bundle url to get the correct response
2. Empty Auth object was used to make api call which was failing on Github redirected address. It will also now avoid using auth(username/password) params if username/password is null.

### DIFF
--- a/stix_shifter/stix_transmission/src/modules/stix_bundle/stix_bundle_connector.py
+++ b/stix_shifter/stix_transmission/src/modules/stix_bundle/stix_bundle_connector.py
@@ -50,7 +50,7 @@ class Connector(BaseConnector):
         if response_code == 200 or response_code == 301:
             return_obj['success'] = True
         else:
-            ErrorResponder.fill_error(return_obj, ['message'])
+            ErrorResponder.fill_error(return_obj, response, ['message'])
         return return_obj
 
     def create_query_connection(self, query):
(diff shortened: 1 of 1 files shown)


stix_shifter/stix_transmission/src/modules/stix_bundle/stix_bundle_connector.py
```diff
@@ -42,7 +42,16 @@ def match(self, pattern, observed_data_sdos, verbose=False):
...
-        return {"success": True}
+        return_obj = dict()
+
+        response = self.call_api('head')
+        response_code = response.status_code
+
+        if response_code == 200 or response_code == 301:
+            return_obj['success'] = True
+        else:
+            ErrorResponder.fill_error(return_obj, ['message'])
```
emishas: you seem to be missing `response` as the second parameter for `ErrorResponder.fill_error`

mdazam1942: right. updated

stix_shifter/stix_transmission/src/modules/stix_bundle/stix_bundle_connector.py
```diff
@@ -42,7 +42,28 @@ def match(self, pattern, observed_data_sdos, verbose=False):
...
+
+        bundle_url = self.connection.get('host')
+        auth = self.configuration.get('auth')
+
+        response = self.call_api(bundle_url, auth, 'head')
+        response_code = response.status_code
+
+        if response_code == 200:
+            return_obj['success'] = True
+        elif response_code == 301:
```
emishas: what I would do here is maybe add default `host` parameter to `ping` and set it to null.
then, when return code is 301, you get the new host value from location and call `ping` again with the new host.

PR #212, opened 2019-11-26
 UT coverage And Host URL change

skairali:


tests/stix_transmission/test_security_advisor.py
```diff
@@ -0,0 +1,244 @@
...
+        "accountID": "abc",
+        "apiKey": "xyz"
+    }
+}
+
+CONNECTION = {
+    "host": "http://test_sec_adv.com/",
+}
+
+import mock
```
mdazam1942: i assume it has no use? need to remove this.

skairali: Done

ajaykumar7-ibm: it's removed

PR #217, opened 2019-12-05
Update Graph QL with more network related fields

skairali:
Update Graph QL with more network related fields


delliott90:
Will this now return more fields in the search results? If you want these new fields to show up in the STIX results, you are going to need to add to your `to_stix_mapy.json`, preferably mapping to official STIX objects.

skairali:
@delliott90  This does not have any change in the return values or STIX Observables.  The underlying graph QL was not searching in certain network details in SA DB. I have included that search.

PR #221, opened 2019-12-13
Doc updates

delliott90:
Updates to add dummy error mappers for the transmission modules, made the overview doc clearer.


OVERVIEW.md
```diff
@@ -75,61 +112,82 @@ You might want to use this library and contribute to development, if any of the
...
 # STIX Pattern:
 "[url:value = 'http://www.testaddress.com'] OR [ipv4-addr:value = '192.168.122.84']"
 ```
 
-##### OUTPUT: Native data source query
+#### OUTPUT: Native data source query
 
 ```
-# Translated Query:
+# Translated Query (using SQL as an example):
```
mdazam1942: would this be AQL?

delliott90: They would be the same. I'm trying to keep the examples generic where possible. This does remind me though that I wanted to clarify the comments in the dummy query _constructo.

PR #225, opened 2019-12-19
Improvement of CLI usage and compatibility 

mdazam1942:
Moved main.py into /stix_shifter/scripts/ in order to use in entry_points that creates stix-shifter executable while installing stix-shifter using pip.

This give better compatibility for people who uses non Linux machine.  Also updated readme to include `Usage` of stix shifter as a script and library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,38 @@ The recommended method for installing the STIX-shifter is via pip.
 pip install stix-shifter
 ```
 
+## Usage
+
+
+### As A Script
+
+The STIX-Shifter comes with a bundled script which you can use to translate STIX Pattern to a native datasource query. It can also be used to translate a JSON data source query result to a STIX bundle of observable objects. You can also send query to a datasource by using a transmission option. 
+
+Details of the command line option can be found [here](https://github.com/opencybersecurityalliance/stix-shifter/blob/master/OVERVIEW.md#how-to-use)
+
+```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -35,6 +35,38 @@ The recommended method for installing the STIX-shifter is via pip.
...
+
+```
+$ stix-shifter translate <MODULE NAME> query "<STIX IDENTITY OBJECT>" "<STIX PATTERN>" "<OPTIONS>"
+```
+Example:
+```
+$ stix-shifter translate qradar query {} "[ipv4-addr:value = '127.0.0.1']" {}
+```
+
+**Note:** In order to create python executable `stix-shifter` from source run the following command from stix-shifter parent directory in your python 3 environment: `python setup.py install`
```
delliott90: This line should probably go right after the first paragraph in this section, since it's a required step before using the CLI. The "Details of the command line" sentence could come here at the end since it's a link to added information. That also brings up the point that the `OVERVIEW` and `adapter-guide` will need to be updated to use this new method of calling the CLI commands, they both contain examples of still calling main.py. Each of those pages should have a line stating that `python setup.py install` will need to be run. Does the setup script need to be installed whenever there are changes?

mdazam1942: right. still some update needs to be done. i just made some quick changes and added WIP tag before i leave for vacation.
About your last question, i need to test that use case but usually when you add new dependency than you may need to create the executable again.

mdazam1942: so i reverted back the main.py in int's original path and allow developer to still use `python main.py ....`
I can add the `Note` part in developer guide as well.

PR #236, opened 2020-01-26
(WIP) Qradar flows

delliott90:
@yurii-klymenko, @mdazam1942 I'd like to discuss what I have here so far, specifically how multiple from-STIX mapping files are getting handled in the stix translator.


stix_shifter/stix_translation/src/modules/base/base_data_mapper.py
```diff
@@ -13,9 +13,10 @@ def fetch_mapping(self, basepath):
         :type basepath: str
         """
         try:
-            filepath = path.abspath(
-                path.join(basepath, "json", "from_stix_map.json"))
-
+            if hasattr(self, 'dialect') and not(self.dialect == 'default'):
+                filepath = self._fetch_from_stix_mapping_file(basepath)
```
yurii-klymenko: it would be a bit straight forward if you pass 'dialect' to the function instead of reading it inside like self.dialect

PR #246, opened 2020-02-25
Adding default ping timeout of 10 seconds

Logicon211:


benjamin-craig:
I'm happy with this as long as Yuri is good.  We did it only on the ping request because we thought some  result calls may last longer than 10 seconds.

stix_shifter/stix_transmission/src/modules/bigfix/bigfix_api_client.py
```diff
@@ -20,7 +21,7 @@ def __init__(self, connection, configuration):
 
     def ping_box(self):
         endpoint = self.endpoint_start + self.PING_ENDPOINT
-        return self.get_api_client().call_api(endpoint, 'GET')
+        return self.get_api_client().call_api(endpoint, 'GET', timeout=DEFAULT_PING_TIMEOUT_IN_SECONDS)
```
yurii-klymenko: here you providing the 'default' value

stix_shifter/stix_transmission/src/modules/utils/RestApiClient.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, host, port=None, cert=None, headers={}, url_modifier_function
         self.url_modifier_function = url_modifier_function
 
     # This method is used to set up an HTTP request and send it to the server
-    def call_api(self, endpoint, method, headers=None, params=[], data=None, urldata=None):
+    def call_api(self, endpoint, method, headers=None, params=[], data=None, urldata=None, timeout=None):
```
yurii-klymenko: but here, the default value is not equal the default value constant, it is puzzling.

delliott90: I think the idea was if any module's api_client didn't pass in the default value, the timeout would be set to None.

yurii-klymenko: it is also an option. In that case 'DEFAULT_PING_TIMEOUT_IN_SECONDS' has to be renamed and to become module local to modules where it is used.

PR #268, opened 2020-03-06
Remove hardcoded modules

yurii-klymenko:


stix_shifter/scripts/stix_shifter.py
```diff
@@ -147,7 +149,24 @@ def main():
 
     args = parent_parser.parse_args()
 
-    if args.command is None:
+    help_and_die = args.command is None
```
mdazam1942: `help_and_exit` ?

yurii-klymenko: has been renamed

stix_shifter_modules/carbonblack/stix_translation/data_mapping.py
```diff
@@ -10,8 +10,8 @@ def __init__(self, options):
         assert "process" in self.map_data and "binary" in self.map_data
 
     def fetch_mapping(self):
-        process_mapping = self._fetch_mapping_file("process_from_stix_map.json")
-        binary_mapping = self._fetch_mapping_file("binary_from_stix_map.json")
+        process_mapping = self._fetch_mapping_file("process_from_stix_map_part.json")
```
mdazam1942: wondering why it is necessary to change the file name? basically you added `_part`

stix_shifter/scripts/stix_shifter.py
mdazam1942: thanks :)

stix_shifter/scripts/stix_shifter.py
```diff
@@ -6,12 +6,13 @@
...
 import time
 from stix_shifter_utils.utils.proxy_host import ProxyHost
+from stix_shifter_utils.utils.module_discovery import module_list, process_dialects
 
 TRANSLATE = 'translate'
 TRANSMIT = 'transmit'
 EXECUTE = 'execute'
 HOST = 'host'
-
+CONNECTOR_MODULES = module_list()
```
mdazam1942: does this need to be global? looks like only usage of this variable is to decide `help_and_exit` flag

stix_shifter_modules/carbonblack/stix_translation/data_mapping.py
yurii-klymenko: carbonblack is the only connector that combines two mapping into the default no name dialect. Dialect discovery logic is based on the filenames, so for the carbonblack it has to find no mapping files for the dialects so only the default dialect is represented for it.

PR #287, opened 2020-03-26
Fix elastic_ecs to_stix_map.json, remove un-needed json.dumps in prox…

freunda:
Removing unneeded json.dumps in translate_results in proxy_host.py, since self.request_args["results"] is already a json


freunda:
A fix for this issue: #288

PR #302, opened 2020-04-06
Add application name to software stix object

mdazam1942:


delliott90:
It looks good. Can you retarget to the develop branch?

PR #305, opened 2020-04-07
Carbon Black mapping fixes

delliott90:


stix_shifter_modules/carbonblack/stix_translation/json/to_stix_map.json
```diff
@@ -69,51 +66,106 @@
       "references": "parent_process"
     }
   ],
+  "parent_id": {
+    "key": "process.x_id",
+    "object": "parent_process"
+  },
```
mdazam1942, line 72: not much familiar with cb events. just wondering, can this `parent_id` object should be set as `parent_ref` in the process `id` object same goes to `unique_id` of process object? 

What I meant is, may be instead setting same value of `process.x_id` in both id and parent_id, can we put a reference?

delliott90: x_id and x_unique_id are custom to CB and different from the PID.  That's why they are in custom properties like that. So CB comes back with `id`, `unique_id`, and `process_pid` for the main process, and `parent_id`, `parent_unique_id`, and `parent_pid` for the parent process

mdazam1942: got it. I was think in terms of general process and parent process referencing.

PR #312, opened 2020-04-20
Input connection and configuration parameter validation 

mdazam1942:


stix_shifter_modules/qradar/configuration/qradar_config.json
```diff
@@ -4,7 +4,8 @@
             "default": "QRadar"
         },
         "host": {
-            "type": "text"
+            "type": "text",
+            "regex": "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
```
yurii-klymenko, line 8: this regex has to applied to all the hosts.

yurii-klymenko: you need to apply regex to name and description fields as well

stix_shifter_modules/qradar/configuration/qradar_config.json
```diff
@@ -40,6 +41,13 @@
         "selfSignedCert": {
             "type": "password",
             "optional": true
+        },
+        "proxy": {
+            "type": "text",
```
yurii-klymenko, line 46: proxy is a section that has to include other fields. https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/qradar/stix_transmission/arielapiclient.py#L32-L33

stix_shifter_utils/utils/param_validator.py
```diff
@@ -0,0 +1,149 @@
...
+    if input_value <= max_value:
+        return True
+    else:
+        return False
+
+def check_regex(input_value, regex_value):
+    pattern = re.compile(regex_value)
+    match_str = pattern.search(input_value)
+
+    if match_str:
```
yurii-klymenko, line 137: return bool(match_str)

stix_shifter_utils/utils/param_validator.py
```diff
@@ -0,0 +1,149 @@
...
+
+def is_leaf(config):
+    if isinstance(config, dict):
+        for key,value in config.items():
+            if isinstance(value, dict):
+                return False
+    return True
+
+def check_min(input_value, min_value):
+    if input_value >= min_value:
```
yurii-klymenko, line 122: it is wrong.
max check is wrong as well

stix_shifter_utils/utils/param_validator.py
```diff
@@ -0,0 +1,149 @@
...
+                    validated_params[key] = input_configs[key]
+                    del input_configs[key]
+                else:
+                    if key not in validated_params:
+                        validated_params[key] = dict()
+                    copy_valid_configs(input_configs[key], expected_configs[key], validated_params[key], errors, key_path)
+                    if not input_configs[key]:
+                        del input_configs[key]
+            else:
+                if not 'optional' in expected_configs[key] and not 'default' in expected_configs[key]:
```
yurii-klymenko, line 107: expect it to add an error even if there is optional param in options but with 'False' value.
107-111 can be replaced with something like that:
if 'default' in expected_configs[key]:
    validated_params[key] = value['default']
elif not ('optional' in expected_configs[key] or expected_configs[key]['optional']):
    errors.append(key_path)

stix_shifter_modules/qradar/configuration/qradar_config.json
mdazam1942: yes I was going to update all config hosts. why do we need to apply regex in name and description? name and description are base config. UI already restricts some characters. and also name and description is not used anywhere in stir-shifter.

stix_shifter_modules/qradar/configuration/qradar_config.json
mdazam1942: yes. will update

stix_shifter_utils/utils/param_validator.py
mdazam1942: sure

stix_shifter_utils/utils/param_validator.py
yurii-klymenko: your conditions validates if there is 'optional' present in the configuration it does not check the value itself.

stix_shifter_modules/qradar/configuration/qradar_config.json
yurii-klymenko: it is a security restriction.

stix_shifter_utils/utils/param_validator.py
yurii-klymenko: https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/base_config.json#L7
min/max checking should work to check string length too.

stix_shifter_modules/qradar/configuration/qradar_config.json
mdazam1942: @yurii-klymenko I understand the security restriction. But do you know what special characters we should allow for name and description? I asked Luke he doesn't have anything validation for name and description. we need to be align with UI in this case. For simplicity, I can just put a regex that allows only string and integer but not sure that's sufficient.

stix_shifter_utils/utils/param_validator.py
mdazam1942: for our use cases, we only specify `optional` when the value is True. I think we had a discussion to remove `"optional": true`. You suggestion may not work. I need to change the conditional check a little bit

stix_shifter_utils/utils/param_validator.py
mdazam1942: so the idea is, if `"type": "text"` check length if `"type": "number"` check value

stix_shifter_modules/qradar/configuration/qradar_config.json
```diff
@@ -40,6 +41,28 @@
...
+                "type": "text",
+                "optional": true
+            },
+            "x_forward_proxy_auth": {
+                "type": "password",
+                "optional": true
+            },
+            "optional": true
+        },
+        "data_lake": {
```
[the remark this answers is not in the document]
mdazam1942: I can put boolean type.
`"data_lake": {
            "type": "boolean",
            "default": false
        }`

stix_shifter_modules/qradar/configuration/qradar_config.json
```diff
@@ -40,6 +41,28 @@
...
+            },
+            "x_forward_proxy": {
+                "type": "text",
+                "optional": true
+            },
+            "x_forward_proxy_auth": {
+                "type": "password",
+                "optional": true
+            },
+            "optional": true
```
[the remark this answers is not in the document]
mdazam1942: right. will remove `"optional": true`

stix_shifter_utils/utils/param_validator.py
```diff
@@ -0,0 +1,162 @@
...
+        if input_value >= min_value:
+            return True
+        else:
+            return False
+
+    if type == 'number':
+        if len(input_value) >= len(min_value):
+            return True
+        else:
+            return False
```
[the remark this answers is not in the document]
mdazam1942: why do we need min value for password type?

stix_shifter_modules/qradar/configuration/qradar_config.json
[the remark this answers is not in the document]
mdazam1942: but that creates another complexity. in current implementation, validator will take it as a required field.

stix_shifter_utils/utils/param_validator.py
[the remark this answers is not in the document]
yurii-klymenko: we do not. we want an exception if there is a max param or any other that we do not validate.

stix_shifter_modules/qradar/configuration/qradar_config.json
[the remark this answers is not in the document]
yurii-klymenko: it makes more sense.

PR #314, opened 2020-04-21
pluggable mapping, refactoring

yurii-klymenko:


stix_shifter/scripts/stix_shifter.py
```diff
@@ -271,15 +279,17 @@ def is_async():
...
             options['stix_validator'] = args.stix_validator
-        if args.data_mapper:
-            options['data_mapper'] = args.data_mapper
         recursion_limit = args.recursion_limit if args.recursion_limit else 1000
         translation = stix_translation.StixTranslation()
         result = translation.translate(
             args.module, args.translate_type, args.data_source, args.data, options=options, recursion_limit=recursion_limit)
+    elif args.command == MAPPING:
+        translation = stix_translation.StixTranslation()        
+        result = translation.translate(args.module, stix_translation.MAPPING, None, None, options=options)
```
delliott90, line 288: Is options always defined by this point?

yurii-klymenko: they are not used. options will be deleted from these calls with next refactoring PR. they will be passed through EntryPoint constructor only.

PR #317, opened 2020-04-23
Updating grammar files for antlr4-python3-runtime 4.8

mdazam1942:


delliott90:
Would you also want to set the minimum version of anltr to 4.8 in the requirements.txt file?

mdazam1942:
> Would you also want to set the minimum version of anltr to 4.8 in the requirements.txt file?

I thought about it but than again we may encounter version conflicts error down the road whenever new antrl version releases because some oasis stix package that we use mostly have the latest antlr.

PR #320, opened 2020-04-28
Set package versions in requirements file

mdazam1942:


yurii-klymenko:
it would be nice to update stix_shifter_modules/cloudsql/requirements.txt as well

PR #326, opened 2020-05-08
Hotfix/3.0.8

yurii-klymenko:
remove urllib usage
prevent sql-injections


delliott90:
I'm getting several unit test failures in test_splunk_json_to_stix.py

yurii-klymenko:
stix2-validator==1.1.2
got a newer version that is not compatible
stix2-validator==2.0.0

PR #331, opened 2020-05-14
Update ToDirectoryPath transformer to remove filename from path value

mdazam1942:


stix_shifter_utils/stix_translation/src/utils/transformers.py
```diff
@@ -4,7 +4,7 @@
 import socket
 import re
 import os
-
+import ntpath
```
yurii-klymenko, line 7: according to https://docs.python.org/3/library/os.path.html ntpath is a Windows specific module. I feel like you have to use more general approach.

mdazam1942: I know. It works with most UNIX and all windows file path conventions. I couldn't find any more generic approach than this. The main reason I went with this approach is, it works with the sample logs where it is currently used.

PR #334, opened 2020-05-21
added timeout in request method calls for searches

mdazam1942:


stix_shifter_modules/azure_sentinel/stix_transmission/api_client.py
```diff
@@ -16,6 +17,11 @@ def __init__(self, connection, configuration):
         auth = configuration.get('auth')
         self.endpoint = '{api_version}/security/alerts'.format(api_version=default_api_version)
         self.host = connection.get('host')
+        options = connection.get('options')
```
yurii-klymenko, line 20: connection is supposed to have options section always since it has default values.

mdazam1942: added this check for the cli

PR #336, opened 2020-05-25
Update base_entry_point.py

freunda:
extract __connector_module from traceback stack using os.path.split to support both Linux and Windows based OS


yurii-klymenko:
fixed the other way in https://github.com/opencybersecurityalliance/stix-shifter/pull/340

PR #337, opened 2020-05-25
Ignore unexpected_params in param_validator.py

freunda:
In contrast to missing_params, unexpected_params should not raise an error (might only log a warning) 
In a scenario of cp4s connected to stix shifter in proxy host mode, there are few unexpected_params that can be ignore:
error occurred: {'unexpected_params': ['description', 'maxConcurrentSearches', 'name', 'requiresProxy', 'resultSizeLimit', 'timeoutLimit', 'use_securegateway', 'wizardType', 'host', 'port', 'type', 'users', 'connectionId', 'description', 'name', 'type']}
. All required parameters exists.


yurii-klymenko:
most of the parameters were included into configuration with default value.

PR #341, opened 2020-05-26
remove unneeded json.dumps

freunda:
remove unneeded json.dumps + print of the results argument in translate_results function. 
Otherwise it will not works when connecting as a proxy connector on cp4s.


mdazam1942:
@freunda going to close this as same changes proposed in another recent pr.

PR #346, opened 2020-06-03
cleanup extra config fields

yurii-klymenko:


mdazam1942:
@yurii-klymenko so we don't need these field anymore in our base config?

yurii-klymenko:
nope, there is a better solution

PR #347, opened 2020-06-05
Fix qradar log source type name property

mdazam1942:


mdazam1942:
> You'll want to remove `devicetypename` from the to-STIX map since it's been removed form the select statement.

yes fixed.

PR #349, opened 2020-06-08
Feature/cb mappings

delliott90:
Update to Carbon Black Response mappings


delliott90:
I created a new issue around the binary mappings https://github.com/opencybersecurityalliance/stix-shifter/issues/355

PR #350, opened 2020-06-09
Adding pythong logging to stdout and stderr

mdazam1942:


yurii-klymenko:
I see a lot of print statements in non test related code, that needs to be fixes.

mdazam1942:
> I see a lot of print statements in non test related code, that needs to be fixes.

@yurii-klymenko pushed the changes. please review

PR #358, opened 2020-06-22
Feature/improve configs

yurii-klymenko:


stix_shifter_modules/lang_en.json
```diff
@@ -4,42 +4,48 @@
...
-            "label": "Query Timeout Limit",
-            "description": "The time limit in minutes for how long the query is run on the data source. The default time limit is 1. When the value is set to zero, there is no timeout. If the query takes longer than 1 min, it is likely to indicate a problem."
+        "options": {
+            "result_limit": {
+                "label": "Result Size Limit",
+                "description": "The maximum number of entries or objects that are returned by search query."
+            },
+            "time_range": {
+                "label": "Query Time Range",
+                "description": "Time range for the search"
```
delliott90, line 21: I wonder if you should add in minutes to this description. Also, should be make it clear it's a time range going back from the current time?

stix_shifter_modules/proxy/configuration/config.json
```diff
@@ -1,20 +1,19 @@
...
-            "default": 443,
-            "type": "number"
+            "type": "number",
+            "default": 443
         },
         "help": {
-            "default": "https://www.ibm.com/support/knowledgecenter/SSTDPP_1.1.0/docs/scp-core/data-sources-proxy.html",
-            "type": "link"
+            "type": "link",
+            "default": "https://www.ibm.com/support/knowledgecenter/SSTDPP_1.1.0/docs/scp-core/data-sources-proxy.html"
```
delliott90, line 16: I guess we have no choice for now, but these links wouldn't be accessible to anyone without an IBM or partner login.

yurii-klymenko: it makes sense to duplicate that page somewhere on a public resource.

Issue #362, opened 2020-07-01
Using backslashes ("\") in STIX values with QRadar backend


barvhaim:
I am trying to execute the following query for cp4s for QRadar 
`select * FROM events WHERE Image LIKE '%\powershell.exe%' limit 1000 START 1593407880000 STOP 1593451080000`

I added custom field at the STIX Shifter and connected to local QRadar, if running on QRadar I do get results, but cp4s won’t let me do the query as `[(process:image_ref.name LIKE '%\powershell.exe')] START t'2020-06-29T05:18:00.000Z' STOP t'2020-06-29T17:18:00.000Z'`, so it pass validation when adding extra \ to be `[(process:image_ref.name LIKE '%\\powershell.exe')] START t'2020-06-29T05:18:00.000Z' STOP t'2020-06-29T17:18:00.000Z'` but then, I don’t get any results since it query for `select * FROM events WHERE Image LIKE '%\\\\powershell.exe%' limit 1000 START 1593407880000 STOP 1593451080000` which does not match the single quote at the image name `%\powershell.exe%`

It worked fine only after adding un-escaping for the query before sending it to QRadar backend
`query = query.replace('\\\\', '\\')`

mdazam1942:
this issue is resolved in [commit](https://github.com/opencybersecurityalliance/stix-shifter/pull/363/commits/8d65e78b17d4424d2bb23b18c453654fb18a5e65)

PR #363, opened 2020-07-07
Extra escaping for the value changing the AQL query value

barvhaim:
I am trying to execute the following query for cp4s for QRadar
`... WHERE qidname LIKE '%event\example%'`

STIX won’t let me do the query as `[(x-ibm-ariel:event_name LIKE '%event\example%')]` so it pass validation when adding extra \ to be `[(x-ibm-ariel:event_name LIKE '%event\\example%')]` but then, I don’t get any results since it query for `... WHERE Image LIKE '%event\\\\example%'` which does not match the single quote at the name '%event\example%'

I pin-point the place where the escaping is added.


barvhaim:
#362 related issue

PR #367, opened 2020-07-08
fix: stix validation failed in artifact base64 object encoding

mdazam1942:


stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
```diff
@@ -50,7 +50,7 @@
     "sourcev6",
     "starttime",
     "endtime",
-    "BASE64(sourcepayload) as base64_source_payload",
-    "BASE64(destinationpayload) as base64_destination_payload"
+    "sourcepayload",
```
delliott90, line 53: Won't these need to use the UTF8 function like in the events fields? I think payload comes in the same for both events and flows. Maybe we should rename these as flowsourcepayload and flowdestinationpayload so we know where they are coming from when looking at the to-stix mapping.

stix_shifter_modules/qradar/stix_translation/json/aql_events_fields.json
```diff
@@ -25,7 +25,6 @@
     "identityhostname",
     "eventcount",
     "PROTOCOLNAME(protocolid) as protocol",
-    "BASE64(payload) as base64_payload",
     "UTF8(payload) as utf8_payload",
```
delliott90, line 28: This could be renamed as `eventpayload` to differentiate it from the flow payloads in the to-stix map.

stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
mdazam1942: right. made that change

stix_shifter_modules/qradar/stix_translation/json/aql_events_fields.json
mdazam1942: right. made that change

stix_shifter_modules/qradar/tests/stix_translation/test_qradar_json_to_stix.py
```diff
@@ -70,14 +70,15 @@ def test_common_prop(self):
 
     def test_cybox_observables(self):
         payload = "Payload"
+        bas64_payload = base64.b64encode(payload.encode('ascii')).decode('ascii')
```
delliott90, line 73: You forgot the `e` in bas64_payload :-)

stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
```diff
@@ -50,7 +50,7 @@
     "sourcev6",
     "starttime",
     "endtime",
-    "BASE64(sourcepayload) as base64_source_payload",
-    "BASE64(destinationpayload) as base64_destination_payload"
+    "sourcepayload as flowsourcepayload",
```
delliott90, line 53: forgot to wrap `sourcepayload` and `destinationpayload` in `UTF8()` function.

stix_shifter_modules/qradar/stix_translation/json/events_from_stix_map.json
```diff
@@ -51,7 +51,7 @@
   },
   "artifact": {
     "fields": {
-      "payload_bin": ["utf8_payload"]
+      "payload_bin": ["eventpayload"]
```
delliott90, line 54: You'll need to add `flowsourcepayload` and `flowdestinationpayload` to the `flows_from_stix_map.json`. I see that originally we had the mapping using the UTF8 representation, but the STIX `payload_bin` property would normally be base64. If someone is using a pattern to search, wouldn't it use base64? We originally had a custom property for the utf8 payload so that it could be searched.

mdazam1942: I think I understand what you are suggesting. so we have implemented this as special case for qradar. though `payload_bin` supposed to be base64 but we expect people to perform text search in this pattern that's why we have utf8_payload mapped to artifact:payload_bin. `eventpayload` still doing the same. 
We can add a custom mapping `x-ibm-ariel:eventpayload` but that will work the same way

mdazam1942: And also will add this mapping for flows

delliott90: Yeah, so for the from-STIX map, `x-ibm-ariel:eventpayload` will map to the UTF8 (human readable) version of the payload and `artifact:payload_bin` will map to the base64 version.

delliott90: Same for flow source and destination payloads.

PR #384, opened 2020-07-27
Feature/mapping in config

yurii-klymenko:


stix_shifter/scripts/stix_shifter.py
```diff
@@ -170,7 +170,7 @@ def main():
 
     help_and_exit = args.command is None
 
-    if args.debug:
+    if 'debug' in args and args.debug:
```
mdazam1942, line 173: is this necessary?

yurii-klymenko: I hit an error on that.

PR #385, opened 2020-07-28
Mapping fix validation

yurii-klymenko:


yurii-klymenko:
@jleemur has to confirm no extra changes are needed.

PR #387, opened 2020-07-29
STIX Shifter that works with CP4S 1.3.0.1 as Proxy Host with QRadar

barvhaim:
Modifications to STIX Shifter to make it work as proxy host mode with QRadar and CP4S 1.3.0.1

### DIFF
--- a/stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix.py
+++ b/stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix.py
@@ -29,4 +29,4 @@ class JSONToStix(BaseResultTranslator):
         except Exception as ex:
             raise TranslationResultException("Error when converting results to STIX: {}".format(ex))
         
-        return results
+        return json.dumps(results, indent=4, sort_keys=False)
--- a/stix_shifter_utils/utils/param_validator.py
+++ b/stix_shifter_utils/utils/param_validator.py
@@ -83,7 +83,7 @@ def param_validator(module, input_configs, start_point=None):
         if isinstance(input_configs, dict):
             error_obj['unexpected_params'] = input_configs
 
-    if error_obj:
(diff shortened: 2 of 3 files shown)


barvhaim:
at file `stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix.py` line 32 `return json.dumps(results, indent=4, sort_keys=False)` - without this change, the result will not appear in CP4S 1.3.0.1, however, this change cause the QRadar test not to pass. @yurii-klymenko should the test be changed? what's your recommendation?

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix.py
```diff
@@ -29,4 +29,4 @@ def translate_results(self, data_source, data):
         except Exception as ex:
             raise TranslationResultException("Error when converting results to STIX: {}".format(ex))
         
-        return results
+        return json.dumps(results, indent=4, sort_keys=False)
```
mdazam1942, line 32: we can't change this back to json strings as that might effect how UDI pipeline handles results. you may need to handle this into the proxy module level.

barvhaim: okay we'll take a look

mdazam1942:
@barvhaim I'm not sure what's the implication on the UDI side of it if we don't handle unexpected_params the way it is now. @yurii-klymenko would need to investigate more into it. since he is in vacation for a week. Is there any way you could handle those unexpected parameters in proxyhost class or even in proxy module? Just wanted to mention, you can make as many changes you want in proxy module or proxyHose if needed.

barvhaim:
since no proxy mode on CP4S 1.4 I think it's not relevant anymore, we'll modify the official connectors

Issue #370, opened 2020-07-13
[QRadar] Escaping `(` and `)` and `"` in value changes AQL meaning


barvhaim:
**Describe the bug**

given the following STIX pattern: `[(process:command_line LIKE 'my command line is (here)')]`, when translating to AQL, the value is escaped and now is `LIKE '%my command line is \(here\)%'` which is different from `'my command line is (here)'` in QRadar.

the AQL `FROM events WHERE ProcessCommandLine LIKE '%my command line is \(here\)%' limit 1000 last 5 minutes` is different than `FROM events WHERE ProcessCommandLine LIKE '%my command line is (here)%' limit 1000 last 5 minutes` and will return zero results.

Using value as `LIKE '%my command line is (here)%'` does seems to be correct with QRadar, I think the escaping inside the value is not needed and changes the AQL.

**To Reproduce**
1. Generate the following event 
`<142>Jul 13 14:37:00 WIN-VM-BAR AgentDevice=WindowsLog	AgentLogFile=Microsoft-Windows-Sysmon/Operational	PluginVersion=7.2.9.105	Source=Microsoft-Windows-Sysmon	Computer=MSEDGEWIN10	OriginatingComputer=MSEDGEWIN10	User=SYSTEM	Domain=NT AUTHORITY	EventID=1	EventIDCode=1	EventType=1	EventCategory=1	RecordNumber=6257	TimeGenerated=1593692483	TimeWritten=1593692483	Level=Informational	Keywords=0x8000000000000000	Task=SysmonTask-SYSMON_CREATE_PROCESS	Opcode=Info Message=IntegrityLevel: System CommandLine: my command line is (here) CurrentDirectory: x Image: example\powershell.exe FileVersion: x ParentImage: C:\Windows\System32\svchost.exe ParentCommandLine: x`

2. Execute using STIX Shifter: `[(process:command_line LIKE 'my command line is (here)')]` you'll get zero results.
3. After removing from `query_constructor.py` line 88 to be `return '{}'.format(value.replace('\"', '\\"'))` and not `return '{}'.format(value.replace('\"', '\\"').replace('(', '\\(').replace(')', '\\)'))`, it will work fine

**Desktop (please complete the following information):**
 - QRadar 7.3.3 Community Edition

barvhaim:
Update - Occur for LIKE clause where value has `"` sign as well, for example: 
[...]

PR #391, opened 2020-07-29
Value escaping in LIKE is unnecessary and changes the value in QRadar

barvhaim:
Following issue #370 

given the following STIX pattern: `[(process:command_line LIKE 'my command line is (here)')]`, when translating to AQL, the value is escaped and now is `LIKE '%my command line is \(here\)%'` which is different from `'my command line is (here)'` in QRadar, and will not return the intended results.

Occurs as well for LIKE clause where value has " sign as well, for example:
`ProcessCommandLine LIKE '%% /C type nul > \"C:\Users\\%\Desktop\\%.exe%' != ProcessCommandLine LIKE '%% /C type nul > "C:\Users\\%\Desktop\\%.exe%'`

### DIFF
--- a/stix_shifter_modules/qradar/stix_translation/query_constructor.py
+++ b/stix_shifter_modules/qradar/stix_translation/query_constructor.py
@@ -79,8 +79,7 @@ class AqlQueryStringPatternTranslator:
 
     @staticmethod
     def _format_like(value) -> str:
-        value = "'%{value}%'".format(value=value)
-        return AqlQueryStringPatternTranslator._escape_value(value)
+        return "'%{value}%'".format(value=value)
 
     @staticmethod
     def _escape_value(value, comparator=None) -> str:


stix_shifter_modules/qradar/stix_translation/query_constructor.py
```diff
@@ -79,8 +79,7 @@ def _format_equality(value) -> str:
 
     @staticmethod
     def _format_like(value) -> str:
-        value = "'%{value}%'".format(value=value)
-        return AqlQueryStringPatternTranslator._escape_value(value)
+        return "'%{value}%'".format(value=value)
```
mdazam1942, line 82: it looks good. I found an existing issue even though I think, it doesn't affect the aql query. If LIKE operator in STIX Pattern contain start and end `%` character than the translated query end up having two `%` character which is unnecessary. We can check the start and end % char and then format the string.

barvhaim: I checked with QRadar, if you use LIKE the extra '%' is ignored ('%%xx%%' treated like '%xx%')

mdazam1942: Right. I'm merging this as it is. if we need we can handle extra % later.

PR #396, opened 2020-08-07
second try

Guy-Galil:


stix_shifter_modules/guardium/stix_translation/data_mapper.py
```diff
@@ -0,0 +1,13 @@
+from os import path
```
mdazam1942, line 1: I don't think data_mapper.py exists anymore in the latest develop and also its not used in the translation module as well.

mdazam1942:
@Guy-Galil  can we close this pr?

PR #414, opened 2020-08-23
Elastic ECS translation with LIKE operator incorrect results

barvhaim:
A query translation for elastic_ecs datasource that contains LIKE operator with a wildcard is behaving incorrectly when there is whitespace between the terms, for example the query "[(x-event:action LIKE 'Registry %')]" which should look for an event action that starts with Registry , will look for matching Registry OR % individually instead. query should result in `event.aciton : "Registry *"` and not `event.aciton : Registry *`


mdazam1942:
@barvhaim can you please update the branch and fix the build. couple of unittest failed.

barvhaim:
@mdazam1942 yes, I'll take a look into

barvhaim:
Update - I found different behavior on several cases, I'll elaborate more in a connected issue,
for now this pull request is not fully fixing the problem.

PR #419, opened 2020-08-27
Missing escaping for ECS str in equality values

barvhaim:
Missing escaping for ECS str in equality values for backslash
pattern 
```
[(windows-registry-key:key = 'HKLM\\SYSTEM\\CurrentControlSet\\Control\\MiniNt' AND x-event:code = 1)] 
```

should be queried as 
```
{
    "queries": [
        "event.code : \"1\" AND registry.key : \"HKLM\\\\SYSTEM\\\\CurrentControlSet\\\\Control\\\\MiniNt\" AND (@timestamp:[\"2020-08-25T11:55:00.000Z\" TO \"2020-08-26T06:05:00.000Z\"])"
    ]
}
```

and not as 
```
{
    "queries": [
        "event.code : \"1\" AND registry.key : \"HKLM\\SYSTEM\\CurrentControlSet\\Control\\MiniNt\" AND (@timestamp:[\"2020-08-25T11:55:00.000Z\" TO \"2020-08-26T06:05:00.000Z\"])"
    ]
}
```


stix_shifter_modules/elastic_ecs/stix_translation/query_constructor.py
```diff
@@ -49,8 +49,10 @@ def _format_set(values) -> str:
 
     @staticmethod
     def _format_equality(value) -> str:
-
-        return '"{}"'.format(value)
+        if isinstance(value, str):
+            return '"{}"'.format(value.replace('\\', '\\\\'))
```
mdazam1942, line 53: @barvhaim kinda confused here. don't we have `_escape_value` function for this? Another question, does double backslash in the string makes an invalid query?

barvhaim: @mdazam1942 The `if-cases` will go to the `_format_equality` case, the `_escape_value` does not get called. it makes incorrect query.

barvhaim:
@mdazam1942 for example key `registry.key` and value `System\CurrentControlSet\Services\W32Time\Config\LastKnownGoodTime`
image


pattern `[(process:command_line = 'foo\\bar')]` translated to `"process.command_line : \"foo\\bar\""`, using this query value format would return zero result - 
image

but if you escape the value backslashes with double backslashes, it will retrieve results - 
image

so pattern should be translated as `[(process:command_line = 'foo\\bar')]` translated to `"process.command_line : \"foo\\\\bar\""`

barvhaim:
I'll reopen with different patch that uses the _escape_value

PR #424, opened 2020-08-28
Fix base64 and host regex

yurii-klymenko:


delliott90:
We shouldn't be merging all of these commits into the release should we?

PR #444, opened 2020-09-22
IPv4 and IPv6 list filter transformers for ECS result `host.ip`

barvhaim:
Given list of IPs from elastic_ecs `host.ip` field, it is IPv4 and IPv6 mixed, so we apply this transformer to separate it to two objects IPv4 and IPv6 
```
{
         "key": "ipv4-addr.value",
         "object": "host_ip_addr_v4",
         "transformer": "FilterIPv4List",
         "unwrap": true
 },
{
         "key": "ipv6-addr.value",
         "object": "host_ip_addr_v6",
         "transformer": "FilterIPv6List",
         "unwrap": true
 },
```


mdazam1942 (commented):
transformer looks ok. @barvhaim I think you haven't pushed any mapping changes to use this transformers

barvhaim:
> transformer looks ok. @barvhaim I think you haven't pushed any mapping changes to use this transformers

it will come on different PR of mappings

mdazam1942 (approved):

PR #446, opened 2020-09-24
Fix deploy script

yurii-klymenko:


deployment/ibm_cloud_pak_for_security/deploy.sh
```diff
@@ -2,6 +2,8 @@
 
 cd "$(dirname "$0")"
 
+PYTHONIOENCODING='utf8'
```
delliott90, line 5: This never gets referenced?

yurii-klymenko: it used by python https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING

delliott90 (approved):

PR #472, opened 2020-10-29
Custom STIX used for mapping fallback added as a cybox object

mdazam1942:


stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -314,8 +314,13 @@ def transform(self, obj):
         # Add unmapped attributes and required property to the observation if it wasn't added via the mapping
         if self.options.get('unmapped_fallback'):
             if self.cust_attributes:
-                self.logger.info('Unmapped fallback is enabled. Adding custom attributes to the obseravble object: {}'.format(self.cust_attributes)) 
-                observation.update({"x-" + self.data_source.lower(): self.cust_attributes})
+                self.logger.info('Unmapped fallback is enabled. Adding custom attributes to the obseravble object: {}'.format(self.cust_attributes))
+                obj = {'type': "x-" + self.data_source.lower()}
+                obj_key = str(len(observation['objects']))
```
delliott90, line 319: Shouldn't this be `len(observation['objects'] + 1` since you are adding one more object to the dictionary?

delliott90: actually never mind, testing it out I can see that's not the case.

delliott90 (approved):

PR #481, opened 2020-11-16
Carbon Black Cloud module

crahan:
## Introduction

I would like to submit this pull request for a [Carbon Black Cloud](https://www.carbonblack.com/products/vmware-carbon-black-cloud-endpoint/) stix-shifter module. I've mapped as many of the fields as possible to STIX and added the remainder of the Carbon Black Cloud fields as custom STIX entries, similar to the current Carbon Black Response stix-shifter module.

## Implementation details

The module is asynchronous and implements all required API endpoints (i.e. ping, delete, query, status, and results). The `ErrorMapper` class leverages the `translation_key` field returned by Cb Cloud and uses it to populate the error code in case of a non-20X status code.

For process executions on Linux, where a process is exec'd (but not forked), details for the exec'd process will be bundled into a single search result. If bash execs zsh then the PID, command line, and username for both processes will be returned together as a list for these fields as part of a single search result. For those cases I've simply converted the list of all PIDs/command lines/usernames to a single string.

As `path` is not used anymore and `parent_name` and `process_name` fields now return the full path to the process binary, I've split that information up into `process_name`/`parent_name` and `process_path`/`parent_path` fields and mapped these to directory STIX objects which are referenced from the appropriate STIX file objects. As hash values for processes and parents are returned as a list with md5 and sha256 values in random order, these have been extracted/converted to `process_md5`/`parent_md5` and `process_sha256`/`parent_sha256` JSON fields as well to allow proper mapping to STIX.

The offset in the `ResultsConnector` class was assumed to start at 0 (which is what Carbon Black Cloud starts with as the first search result). If this assumption is incorrect (i.e. results offset count from 1 in stix-shifter) I'll update accordingly.

Finally, Carbon Black Cloud does not differentiate between process and binary searches like Carbon Black Response does. Binary data is now queryable through the process search API. As a result, there are no dialect definitions in this module.

Currently unit tests are missing as I am not familiar with the proper structure/definition on how these should be written. I can add these based on feedback.

## Translation screenshots

(Please note: screenshots are not necessarily related to the same search/query)

### Carbon Black Cloud JSON result to STIX
99200653-d8c86c00-27a6-11eb-93d1-c77d5485e4e3
99200807-d4508300-27a7-11eb-9eb8-3143a7cecde6

### STIX to Carbon Black Cloud query string
Screen Shot 2020-11-16 at 00 59 38

### Data source ping
ping

### Query submission
query submit

### Query status
query status

### Query delete
query delete

### Query results
query results


stix_shifter_modules/cbcloud/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,36 @@
...
+from stix_shifter_utils.utils import logger
+
+
+class ResultsConnector(BaseResultsConnector):
+    def __init__(self, api_client):
+        self.api_client = api_client
+        self.logger = logger.set_logger(__name__)
+
+    def create_results_connection(self, search_id, offset, length):
+        # TODO: confirm offset starts at 0 or 1 (CbCloud starts at 0)
```
mdazam1942, line 14: I think it starts at 0. And length value I the number of results

stix_shifter_modules/cbcloud/stix_transmission/error_mapper.py
```diff
@@ -0,0 +1,24 @@
...
+    DEFAULT_ERROR = ErrorCode.TRANSMISSION_MODULE_DEFAULT_ERROR
+    logger = logger.set_logger(__name__)
+
+    @classmethod
+    def set_error_code(cls, json_data, return_obj):
+        error_code = cls.DEFAULT_ERROR
+
+        try:
+            # Carbon Black Cloud stores error messages in 'translation_key'
+            error_code = json_data['translation_key']
```
mdazam1942, line 17: is `translation_key` object contains the error codes? if you have the list of error codes you can create a dictionary and map the error codes just like we have in other modules.

stix_shifter_modules/cbcloud/tests/stix_translation/test_cbcloud_json_to_stix.py
```diff
@@ -0,0 +1 @@
+# TODO
```
mdazam1942, line 1: you can take a look at other modules how the tests are written. there are not much difference while writing unit tests for translations

mdazam1942 (changes requested):
@crahan let us know if have questions about any of the comments. we can help you writing the unitests if you need

stix_shifter_modules/cbcloud/configuration/lang_en.json
```diff
@@ -0,0 +1,37 @@
...
+        },
+        "sni": {
+            "label": "Server Name Indicator",
+            "description": "The Server Name Indicator (SNI) enables a separate hostname to be provided for SSL authentication."
+        }
+    },
+    "configuration": {
+        "auth": {
+            "orgkey": {
+                "label": "Org key",
```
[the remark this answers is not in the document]
crahan: Org Key is the name the platform uses to refer to this field.

stix_shifter_modules/cbcloud/configuration/lang_en.json
```diff
@@ -0,0 +1,37 @@
...
+                "description": "Organization key"
+            },
+            "password": {
+                "label": "Token",
```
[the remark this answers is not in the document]
crahan: Similar to Org Key, I've used the same field name that the platform uses to refer to this value and added a description with the full name to clarify. I can remove the description of course, but that would make the field names different in the connector and the instance.

stix_shifter_modules/cbcloud/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,180 @@
...
+      "object": "process",
+      "references": "process_file"
+    }
+  ],
+  "process_pid": {
+    "key": "process.pid",
+    "object": "process"
+  },
+  "process_guid": {
+    "key": "process.x_unique_id",
```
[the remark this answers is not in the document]
crahan: Yes, that can be remapped.

stix_shifter_modules/cbcloud/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,273 @@
...
+        stop = cls._stix_to_cbcloud_timestamp(qualifier.stop)
+        return f'({expression}) AND device_timestamp:[{start} TO {stop}]'
+
+    @staticmethod
+    def _escape_value(value) -> str:
+        """Trim whitespace and escape specific characters in a value."""
+        if isinstance(value, str):
+            value = value.strip()
+        if isinstance(value, str) and not re.match(r'^\[[\*\sa-z0-9.:-]*to[\*\sa-z0-9.:-]*\]$', value.lower()):
+            value = value.replace('\\', '\\\\')
```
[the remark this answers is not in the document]
crahan: We did some testing with Danny Elliott and there was some weird behavior when testing this. This is set up the same as the  `carbonblack` connector (https://github.com/opencybersecurityalliance/stix-shifter/blob/a1296f9b51618dd0b4f432ec08e9469b3548722d/stix_shifter_modules/carbonblack/stix_translation/query_constructor.py#L70)

stix_shifter_modules/cbcloud/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,273 @@
...
+            value = value.replace('\"', '\\"')
+            value = value.replace('(', '\\(')
+            value = value.replace(')', '\\)')
+            value = value.replace(' ', '\\ ')
```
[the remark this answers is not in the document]
crahan: Changes a space into a `\ `. Carbon Black Cloud (ThreatHunter) requires a certain characters to be backslash-escaped when submitting a query that is not enclosed in quotes.

stix_shifter_modules/cbcloud/stix_transmission/error_mapper.py
crahan: I do not have a mapping for these unfortunately.

stix_shifter_modules/cbcloud/configuration/lang_en.json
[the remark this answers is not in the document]
mdazam1942: in that case, you can keep the Label and description as it is.

stix_shifter_modules/cbcloud/configuration/lang_en.json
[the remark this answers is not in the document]
mdazam1942: in that case, you can keep the Label and description as it is.

stix_shifter_modules/cbcloud/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,180 @@
+{
+  "process_username": [
+    {
+      "key": "user-account.account_login",
```
mdazam1942, line 4: so the `user-account` object also needs `user_id`. as per stix specification because that's a required property. I wonder if you can set it to user_id instead or if you have any other field that can be set as user_id

stix_shifter_modules/cbcloud/stix_transmission/status_connector.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Construct a response object
+        return_obj = dict()
+        return_obj['success'] = False
+
+        if response_dict and response_code == 200:
+            completed = response_dict['completed']
+            contacted = response_dict['contacted']
+            return_obj['success'] = True
+            return_obj['progress'] = {
```
mdazam1942, line 60: nested json for `progress` property will throw error to the caller as the caller of the status endpoint is expecting just the progress number. you can do the following instead-

```
return_obj['progress'] = completed
return_obj['contacted'] = contacted
```

stix_shifter_modules/cbcloud/stix_translation/json/to_stix_map.json
crahan: Will change. I was using like wat appeared to be the most appropriate field (a string/word seemed to fit better with account_login than with id, which sounded like it would be more of a int/number). If the stix specification requires at the very least to have 'id' set in a user_account, then I can change. Cb Cloud does not have a user ID field, just process_username.

stix_shifter_modules/cbcloud/stix_transmission/status_connector.py
crahan: Perfect, that's an easy fix.

crahan: This has been updated per the suggestion.

stix_shifter_modules/cbcloud/stix_translation/json/to_stix_map.json
crahan: This has been updated together with the equivalent value in the `from_stix_map.json` file.

stix_shifter_modules/cbcloud/stix_transmission/results_connector.py
crahan: Comment removed as functioning has been clarified.

stix_shifter_modules/cbcloud/configuration/lang_en.json
```diff
@@ -0,0 +1,37 @@
+{
+    "connection": {
+        "host": {
+            "label": "Carbon Black Cloud instance hostname",
```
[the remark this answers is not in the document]
crahan: This has been updated per the recommendation.

stix_shifter_modules/cbcloud/configuration/config.json
```diff
@@ -0,0 +1,38 @@
...
+        },
+        "help": {
+            "type": "link",
+            "default": "https://add.a.link.to.the.carbonblackcloud/connector/documentation.html"
+        },
+        "sni": {
+            "type": "text",
+            "optional": true
+        },
+        "selfSignedCert": {
```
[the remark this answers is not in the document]
crahan: `selfSignedCert` and `sni` have both been removed.

stix_shifter_modules/cbcloud/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,180 @@
+{
```
mdazam1942, line 1: we found another important issue. two required stix objects are missing in the mapping `first_observed` and `last_observed`. I found few timestamp field in the results such as `backend_timestamp`, `device_timestamp` and `process_start_time`. one of them may needs to be mapped to those fields.

stix_shifter_modules/cbcloud/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,273 @@
...
+        ComparisonExpressionOperators.And: 'AND',
+        ComparisonExpressionOperators.Or: 'OR',
+        ComparisonComparators.Equal: ':',
+        ComparisonComparators.NotEqual: ':',
+        ComparisonComparators.GreaterThan: ':',
+        ComparisonComparators.GreaterThanOrEqual: ':',
+        ComparisonComparators.LessThan: ':',
+        ComparisonComparators.LessThanOrEqual: ':',
+        ObservationOperators.Or: 'OR',
+        ObservationOperators.And: 'AND'
```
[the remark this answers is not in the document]
crahan: It does not. AND, OR, and NOT are all that's available.

mdazam1942 (approved):

PR #484, opened 2020-11-18
Cb timerange fix

delliott90:


stix_shifter/scripts/stix_shifter.py
```diff
@@ -264,8 +264,8 @@ def is_async():
 
         transmission = stix_transmission.StixTransmission(args.transmission_module, connection_dict, configuration_dict)
         results = []
-        if not 'queries' in dsl:
-            print(json.dumps(dsl, indent=4))
+        print(json.dumps(dsl, indent=4))
```
mdazam1942, line 267: use logger: `log.info('Translated Queries: \n' + json.dumps(dsl, indent=4))`

and also I just found line 265 and 263 are exactly the same calling transmission. I think we should remove one of them:
`transmission = stix_transmission.StixTransmission(args.transmission_module, connection_dict, configuration_dict)`

mdazam1942 (approved):
recommended to fix what I posted in the comment but merging is safe even without that

PR #485, opened 2020-11-18
stix_bundle: Drop duplicate matches from cti-pattern-matcher

pcoccoli:
The stix_bundle connector is returning duplicate results when the bundle contains observations with number_observed > 1.  The root cause is here in cti-pattern-matcher: https://github.com/oasis-open/cti-pattern-matcher/blob/bcd37eeb8c44e012ff1dbe4434c510f343575840/stix2matcher/matcher.py#L1006

However, since the pattern matcher really only needs to return true or false, I can't say that duplicating the SDO in there is a bug.  I would argue that the bundle connector needs to drop those duplicates.

This commit drops a match if it's already in the list of matching SDOs.  There is likely a better/more efficient way to do this; this change is quick and dirty.

### DIFF
--- a/stix_shifter_modules/stix_bundle/stix_transmission/connector.py
+++ b/stix_shifter_modules/stix_bundle/stix_transmission/connector.py
@@ -37,7 +37,8 @@ class Connector(BaseSyncConnector):
         if found_bindings:
             matching_sdos = []
             for binding in found_bindings:
-                matching_sdos = matching_sdos + matcher.get_sdos_from_binding(binding)
+                matches = [match for match in matcher.get_sdos_from_binding(binding) if match not in matching_sdos]
+                matching_sdos.extend(matches)
         else:
             matching_sdos = []
 


pcoccoli:
To reproduce the issue:
```
$ stix-shifter transmit stix_bundle '{"host": "https://raw.github.com/opencybersecurityalliance/stix-shifter/master/data/cybox/qradar/qradar_observed_2000.json"}' '{"auth":""}' results "[ipv4-addr:value = '127.0.0.1']" 0 10000 | jq . | grep '"id"' | sort | uniq -c | sort -nr | head
     21       "id": "observed-data--cfaa4a35-7381-432d-81f6-1695985d7464",
     21       "id": "observed-data--3659300c-0e79-4cd5-ba57-d5215d1a2e1d",
     10       "id": "observed-data--1b2a38d0-c12c-4ef9-8073-53c189ca6e28",
      2       "id": "observed-data--f9f23d0e-3727-4828-8461-d01010409f1c",
      2       "id": "observed-data--9955e544-24f1-4a82-b6dd-4ba541ba35df",
      1       "id": "observed-data--ffedb26d-2cce-49d4-9c7c-de4396368957",
      1       "id": "observed-data--ffdf6355-2239-48d1-a912-d58f8cb54f04",
      1       "id": "observed-data--ffd018b9-8da0-4557-b66d-ed901f164e3d",
      1       "id": "observed-data--ffa1c45c-34fb-49e1-a6ff-39b7633f6da6",
      1       "id": "observed-data--ff72e68d-96aa-4043-855e-b387753fc6a3",
```

After the fix, every `id` should have a count of 1.

mdazam1942 (approved):

PR #493, opened 2020-11-23
Investigator sysmon and windows

aviv1ron1:
## x-ibm-host object
Adds a new object for describing the host related to events.
similar across qradar and elastic_ecs modules.

| property name | type | description |
|--|--|--|
| type | `string` | x-ibm-host |
|hostname|`string`|name of this host|
|ip_refs|`list` of type `object-ref`| references the ip addresses related to this host. must be `ipv4-addr` or `ipv6-addr`|
|mac_refs|`list` of type `object-ref`| references the mac addresses related to this host. must be of type `mac-addr`|

Example:

    {
	    "0":
	    {
	        "type": "ipv4-addr",
	        "value": "9.9.9.9"
	    },
	    "1":
	    {
	        "type": "x-ibm-host",
	        "ip_refs": ["0", "2"],
	        "hostname": "example host",
	        "mac_refs": ["3"]
	    },
	    "2":
	    {
	        "type": "ipv6-addr",
	        "value": "ffff:1111:ffff:1111:ffff:1111:ffff:1111",
	        "resolves_to_refs": ["7"]
	    },
	    "3":
	    {
	        "type": "mac-addr",
	        "value": "00-00-11-00-22-00"
	    }
    }

## x-ibm-event
Adds a new object for describing the events.
Similar across qradar and elastic_ecs modules.
Definition follows closely to the elastic ecs event object.

| property name | type | description |
|--|--|--|
| type | `string` | x-ibm-event |
| action|`string`|describes the event action. for example process create|
|category|`list` of type `string`|categories describing the event. for example "process" or "system"|
|code|`integer`|event code or event id from the original log source|
|created|`timestamp`|when this event was created|
|module|`string`|Name of the module this data is coming from. for example "sysmon"|
|original_ref|`object-ref`|references the original raw event data. must be of type `artifact`|
|outcome|`string`|the outcome of this event. for example "process create success"|
|provider|`string`|the provider of this log. for example "Microsoft Windows Security Event Log"|
|agent|`string`|describes the name of the agent that collected this event log|
|host_ref|`object-ref`|references the host related to this event. must be of type `x-ibm-host`|
|url_ref|`object-ref`|references a url related to this event. must be of type `url`|
|file_ref|`object-ref`|references a file related to this event. must be of type `file`|
|process_ref|`object-ref`|references a process related to this event. must be of type `process`|
|parent_process_ref|`object-ref`|references a parent process related to this event. must be of type `process`|
|domain_ref|`object-ref`|references a domain related to this event. must be of type `domain-name`|
|domain_ref|`object-ref`|references a domain related to this event. must be of type `domain-name`|
|registry_ref|`object-ref`|references windows registry data related to this event. must be of type `windows-registry-key`|
|network_ref|`object-ref`|references network traffic related to this event. must be of type `network-traffic`|
|ip_refs|`list` of type `object-ref`|references any ip addresses related to this event. must be of type `ipv4-address` or `ipv6-address`|
|user_ref|`object-ref`|references the user related to this event. must be of type `user-account`|

In addition this object may contain any field defined in the elastic ecs event definition, but string values are not obligated to be following any schema and may differ between elastic and qradar data sources.

Example (references omitted for simplicity):

    {
        "type": "x-ibm-event",
        "action": "Process Create",
        "outcome": "Process Creation Success",
        "host_ref": "3",
        "created": "1970-01-18T17:19:29.112Z",
        "provider": "Microsoft Windows Security Event Log",
        "url_ref": "8",
        "file_ref": "9",
        "process_ref": "13",
        "parent_process_ref": "16",
        "category": ["System"],
        "original_ref": "17",
        "agent": "WindowsAuthServer@XXXX",
        "user_ref": "18",
        "domain_ref": "19",
        "registry_ref": "20"
    }

## process object

Adds support for `process` object as defined by STIX 2.0 spec for both qradar and elastic_ecs
Example:

	{
	    "0":
	    {
	        "type": "process",
	        "pid": 1248,
	        "binary_ref": "1",
	        "command_line": "C:\\WINDOWS\\system32\\svchost.exe -k DcomLaunch -p",
	        "name": "svchost.exe"
	    },
	    "1":
	    {
	        "type": "file",
	        "name": "svchost.exe",
	        "parent_directory_ref": "2"
	    },
	    "2":
	    {
	        "type": "directory",
	        "path": "C:\\Windows\\System32"
	    }
	}

## registry object
Adds support for `windows-registry-key` as defined by STIX 2.0 spec for both qradar and elastic_ecs.
elastic_ecs module might have additional registry related fields in `x-ecs-registry`
Example:

    {
        "type": "windows-registry-key",
        "key": "HKEY_LOCAL_MACHINE\\a\\b\\c",
        "values": [
        {
            "name": "val"
        }]
    }

## dns-ext extension
Adds an extension to the `network-traffic` object to describe dns queries and answers. The key for this extension when used in the **extensions** dictionary **MUST** be dns-ext.
This extension mostly follows the elastic_ecs dns object and aligns between elastic_ecs module to qradar module.
| property name | type | description |
|--|--|--|
| question.name_ref | `object-ref` | references the dns query domain name in question. must be of type `domain-name` |
|resolved_ip_refs|`list` of type `object-ref`| references ip addresses that were resolved by the answer. must be of type `ipv4-address` or `ipv6-address`|

In addition this object might contain any fields defined in the elastic ecs dns schema.
Example:

    {
      "0":
      {
          "type": "network-traffic",
          "extensions":
          {
              "dns-ext":
              {
                  "question":
                  {
                      "name_ref": "1"
                  },
                  "resolved_ip_refs": ["2", "3"]
              }
          }
      },
      "1":
      {
          "type": "domain-name",
          "value": "domain.com"
      },
      "2":
      {
          "type": "ipv4-addr",
          "value": "9.9.9.9"
      },
      "3":
      {
          "type": "ipv6-addr",
          "value": "aaaa::bbbb:1111:2222:3333"
      }
    }


delliott90:
I see that some properties for the new objects as still mapped to the old custom objects as well. So `x-qradar.high_level_category_name` and `x-ibm-event.category` for example. Typically, we would just move them over to the new object so we don't have duplicate data in the results. The same goes for the from-STIX pattern mapping. That's not to say I can't be convinced otherwise.  I'm not sure what the STIX convention / best practice is for mapping duplicate data in separate properties. @JasonKeirstead may have some thoughts.

mdazam1942 (approved):

delliott90 (approved):

JasonKeirstead:
Why do we need this "host" when we have domain-name already

pcoccoli:
We need something to indicate where a sysmon event happened (this is actually needed for all endpoint events, including EDR systems).  I assume this is what the custom host object represents?  Seems like it could have been simpler, like a single `x_observed_on_ref` reference pointing to a domain-name or ipv4-addr or ...
It also seems strange call these `x-ibm-...` as they're not specific to IBM.  Why not `x-oca-...`?

JasonKeirstead:
So is this referring to an asset? We should probably use "asset" as "where it is happening" could be another type of asset than a host.

dennispo:
Depending on the data source, host, is not always represented by the `domain-name` field. A very common example is where a host is represented by IP address. For a containerized environment it may be a container-name, worker-node, etc...

pcoccoli:
It would be great if x-ibm-host (or x-oca-asset or whatever it winds up being) was also used with MSATP, CbR, Splunk, and whatever other data sources have such a concept (probably all of them?).

JasonKeirstead:
@dennispo Right so that means it should not be a host. a container is not a host. its an asset though.

dennispo:
@JasonKeirstead I don't have a strong opinion on the naming, just wanted to claim that this entity is more generic than a `domain-name`. Another interesting question is: how to identify hosts/assets across different reports (e.g. once reported by IP or MAC address and another time reported using domain-name)?

dennispo:
> It would be great if x-ibm-host (or x-oca-asset or whatever it winds up being) was also used with MSATP, CbR, Splunk, and whatever other data sources have such a concept (probably all of them?).

This pull requests contains support for QRadar and Elastic. We are working to add Splunk support as well. It will actually support all the custom objects: `x-ibm-host`, `x-ibm-event` and  `dns-ext`

Neo23x0:
Please explain the the `ibm` in `x-ibm-host`. 
I don't get it. What does the hostname have to do with IBM?

JasonKeirstead:
@Neo23x0 The x-foo- decorator is a best practice for STIX 2 custom objects. 

I agree it *should not* say IBM anything.  It should be x-oca, at minimum....

Neo23x0:
I've received a pull request for our project, but it seems that this pull request hasn't been merged yet.
At first glance it looked like someone from IBM tried to hijack our config 😄 

Screenshot 2021-01-05 at 19 08 53

That change from a generic looking name to some field name that contains "ibm" is odd.

JasonKeirstead:
@Neo23x0 The guideline is

"Custom Property names SHOULD start with “x_” followed by a source unique identifier (such as a domain name with dots replaced by underscores), an underscore and then the name. For example, x_example_com_customfield."

It is to assist someone looking at STIX content to know who they should seek out as to what those fields mean in the data model. This was before the currently-proposed more formal extension process in STIX 2.1 existed.

My main concern is that this should not say IBM, as STIX Shifter is not an IBM project. It does not matter if the change came from IBM or not, that's not relevant.

delliott90:
I thought the reason we included ibm in `x-ibm-finding` was because that's where the newly proposed object was coming from. How would the host object be any different?

PR #500, opened 2020-12-07
PR for Amazon Athena - Fix for exception scenarios

annish-offl:
Fix for the exception raised with invalid amazon athena service params

### DIFF
--- a/stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
+++ b/stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
@@ -2,8 +2,7 @@ from stix_shifter_utils.modules.base.stix_transmission.base_connector import Bas
 from stix_shifter_utils.utils.error_response import ErrorResponder
 import json
 import re
-import random
-import string
+import uuid
 
 # static lookup for athena service types
 service_types = {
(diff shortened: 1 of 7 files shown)


stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
```diff
@@ -35,10 +37,14 @@ def create_query_connection(self, query):
                 query = json.loads(query)
             query_service_type = list(query.keys())[0]
             config_details = service_types[query_service_type]
+            if all([True if config not in self.connection.keys() else False for config in config_details]):
+                random_id = ''.join([random.choice(string.ascii_letters + string.digits) for _ in range(30)])
```
mdazam1942, line 41: why do you need to generate id like this? can't you just generate a random UUID?

stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
```diff
@@ -35,10 +37,14 @@ def create_query_connection(self, query):
...
+                return_obj = {'success': True, 'search_id': 'athena-' + random_id.lower() + ':' + query_service_type}
```
mdazam1942, line 42: I would make the id more sense like add dummy in the search id instead of Athena such as `a9afe0ac-b6cf-4569-a4eb-730bd6cdfd5e-dummy:guardduty`

stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
annish-offl: @mdazam1942 The code change for generating UUID using the python package is done and pushed to PR branch.

stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
annish-offl: @mdazam1942 The "Athena" prefix is now removed and the dummy search ID is suffixed with the string "dummy". The changes are updated to PR branch.

annish-offl:
Commit  #2b36b74 is a minor renaming for the data source field "id" to "finding id", to ensure there is no ambiguity between the "id" attribute of STIX bundle and the fields of unmapped custom object, due to the STIX format validation.

mdazam1942 (approved):

PR #506, opened 2020-12-16
fix language param

yurii-klymenko:


delliott90 (approved):

stix_shifter/stix_translation/stix_translation.py
```diff
@@ -86,7 +88,7 @@ def translate(self, module, translate_type, data_source, data, options={}, recur
                     dialects_used = 0
                     for dialect in dialects:
                         query_translator = entry_point.get_query_translator(dialect)
-                        if language == query_translator.get_language():
+                        if not language or language == query_translator.get_language():
```
mdazam1942, line 91: is there any possibility of language being None?

stix_shifter_modules/qradar/entry_point.py
```diff
@@ -11,4 +11,4 @@ def __init__(self, connection={}, configuration={}, options={}):
         else:
             self.setup_translation_simple(dialect_default='flows')
             dialect = 'aql'
-            self.add_dialect(dialect,  AqlQueryTranslator(options, dialect, None), default_include=False, default=True)
+            self.add_dialect(dialect,  AqlQueryTranslator(options, dialect, None), default_include=True)
```
mdazam1942, line 14: wondering what is the difference between default and default_include and how does it work after the entry point?

stix_shifter_modules/qradar/stix_translation/aql_query_translator.py
```diff
@@ -38,28 +38,30 @@ def parse_query(self, data):
...
-                result[labels[label]] = self.search_for_pattern(data, time_patterns, label)
+                result[labels[label]] = self.__search_for_pattern(data, time_patterns, label)
 
         if not result.get(END_TIME):
             current_time = datetime.now()
             result[END_TIME] = int(float(datetime.strptime(str(current_time), '%Y-%m-%d %H:%M:%S.%f').strftime('%s.%f'))*1000)
 
         return result
 
+    def transform_query(self, data):
```
mdazam1942, line 49: is there any reason `transform_query`  need to overriding?

stix_shifter/stix_translation/stix_translation.py
yurii-klymenko: it is possible, check line 74. if module is provided explicitly then the default language value has to be ignored for the sake of the provided module.
for example:
`it is possible, check line 74. if module is provided explicitly then the default language value has to be ignored for the sake of the provided module.`

stix_shifter_modules/qradar/entry_point.py
yurii-klymenko: one dialect per language has to have 'default=True' to be used for 'parse' calls since for calls like:
`python main.py translate qradar parse '{}' "[url:value = 'https://google.com/']" '{"language":"stix"}'`
dialect is not specified in cli. parsing will be processed with dialect='flows' since it is the default one for 'stix' language.

stix_shifter_modules/qradar/stix_translation/aql_query_translator.py
yurii-klymenko: parent's transform_query does antlr parsing for 'stix' language, it is not needed for 'aql' dialect.

mdazam1942 (approved):

PR #509, opened 2021-01-04
Gmt

Guy-Galil:
Per SECCON-8815 
Added code so if Guardium supports TimeZone conversion passes inputTZ=GMT to Guardium.
In this case on the Guardium appliance query from_date and to_date are converted from GMT to the appliance's time zone, dates it results are converted from appliance timezone to GMT.
As this functionality was not included in Guardium 11.3 the connector's code recognizes if inputTZ is not supported and sends the query without it.
Similar fix will be added in CAR soon


stix_shifter_modules/guardium/stix_transmission/guard_utils.py
```diff
@@ -113,11 +114,28 @@ def handle_report(self, report_name, params, index_from, fetch_size):
...
         results = ""
         #context().logger.debug('-------------------  ' + report_name + ' ----------------------')
-        params_set = {"reportName":"{0}".format(report_name), "indexFrom": "{0}".format(index_from), "fetchSize": "{0}".format(fetch_size), "reportParameter":params}
+        params_set = {"reportName":"{0}".format(report_name), "indexFrom": "{0}".format(index_from), "fetchSize": "{0}".format(fetch_size), "reportParameter":params, "inputTZ":"GMT"}
         json_dump = json.dumps(params_set)
         rest_data = str(json.loads(json_dump))
-
-        response = requests.post(self.url+self.report_target, data=rest_data, headers=self.headers, verify=False)
+        # For compatibility with Guardium - inputTZ parameter was aded after v11.3 so in case it does not exist execute the query without it
+        for x in range(2):
```
mdazam1942, line 121: not sure if this is the right way. you can just send the request with inputTZ first and check the response if error occurs update the the param and then send the request again. I would remove the for loop for this case

Guy-Galil: This is what I actually do.
The performs the second round only if error code is 27

stix_shifter_modules/guardium/stix_transmission/guard_utils.py
```diff
@@ -113,11 +114,28 @@ def handle_report(self, report_name, params, index_from, fetch_size):
...
+                    if errorCode ==  "27":
+                        params_set.pop("inputTZ")
+                        json_dump = json.dumps(params_set)
+                        rest_data = str(json.loads(json_dump))
+                        self.logger.warn("InputTZ not suppoerted - running query without it")
+                        print("InputTZ not suppoerted - running query without it")
+                        continue
+                except:
+                    pass    
+            break
```
Guy-Galil, line 138: See this block of code - 
In case error is 27 (no api function found for the parameters passed) inputTZ is removed (pop) from param_set and second iteration of the loop is performed.
Otherwise the loop is exited (break)

stix_shifter_modules/guardium/stix_transmission/guard_utils.py
```diff
@@ -113,11 +114,28 @@ def handle_report(self, report_name, params, index_from, fetch_size):
...
+                    errorCode = results["ErrorCode"]
+                    print(errorCode)
+                    if errorCode ==  "27":
+                        params_set.pop("inputTZ")
+                        json_dump = json.dumps(params_set)
+                        rest_data = str(json.loads(json_dump))
+                        self.logger.warn("InputTZ not suppoerted - running query without it")
+                        print("InputTZ not suppoerted - running query without it")
+                        continue
+                except:
```
mdazam1942, line 136: not sure what's the point of having this try/except if you are not throwing any exception.

stix_shifter_modules/guardium/stix_transmission/guard_utils.py
mdazam1942: I'm actually talking about removing the for loop instead just send the request and check the response and then again send the request based on the response. you don't need to for loop for that. 

```
        response = requests.post(self.url+self.report_target, data=rest_data, headers=self.headers, verify=False)
        results = response.json()
        check_list = isinstance(results, list)
        if check_list == False:
            errorCode = results["ErrorCode"]
            if errorCode ==  "27":
                params_set.pop("inputTZ")
                json_dump = json.dumps(params_set)
                rest_data = str(json.loads(json_dump))
                self.logger.warn("InputTZ not suppoerted - running query without it")
                response = requests.post(self.url+self.report_target, data=rest_data, headers=self.headers, verify=False)
```

Guy-Galil: This will work too.
I started this on CAR connector where parsing the results is more complex and I did not want to copy/paste the same code so I did the loop.
Will change it here for clarity

stix_shifter_modules/guardium/stix_transmission/guard_utils.py
Guy-Galil: I am not sure if exceptions work the same way in Python - in Java this would prevent the exception from being thrown upwards.

stix_shifter_modules/guardium/stix_transmission/guard_utils.py
Guy-Galil: I committed a modified code per your comment.

mdazam1942 (approved):

PR #510, opened 2021-01-07
Remove codes for binary searches

mdazam1942:


delliott90 (approved):
Ran a small execute test with `"[process:name = 'svchost.exe']  START t'2021-01-04T00:00:00.000Z' STOP t'2021-01-04T01:00:00.000Z'"` and got the same results on both this branch and develop.

PR #523, opened 2021-01-18
remove RegistryValueData from qradar mappings

aviv1ron1:
This commit removes the mappings of the field "Registry Value Data" from Qradar mappings.
This field is redundant, it is not mapped to any stix object. It is mostly null since it is not being parsed by Qradar DSM.
We will work to investigate how to re-introduce this field in the future, with a correct mapping after we make sure that Qradar populates it correctly.


aviv1ron1:
@mdazam1942

mdazam1942 (approved):

PR #565, opened 2021-03-18
Fix results stix remote call

yurii-klymenko:


stix_shifter/stix_transmission/stix_transmission.py
```diff
@@ -60,10 +60,14 @@ def results(self, search_id, offset, length):
             return return_obj
 
     def results_stix(self, search_id, offset, length, data_source):
-        result = self.results(search_id, offset, length)
-        if result and 'success' in result and result['success']:
-            result = self.entry_point.translate_results(data_source, json.dumps(result['data']))
-        return result
+        try:
+            if self.init_error is not None:
```
delliott90, line 64: Couldn't this just be `if self.init_error`?

delliott90 (approved):

PR #567, opened 2021-03-23
Consistent language and formatting

benjamin-craig:
Consistent language and formatting

### DIFF
--- a/stix_shifter_modules/lang_en.json
+++ b/stix_shifter_modules/lang_en.json
@@ -5,7 +5,7 @@
         },
         "name": {
             "label": "Data source name",
-            "description": "Assign a name to identify the connection minimum of 5 characters"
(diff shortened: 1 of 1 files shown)


stix_shifter_modules/lang_en.json
```diff
@@ -5,11 +5,11 @@
         },
         "name": {
             "label": "Data source name",
-            "placeholder": "Add a data source name"
+            "description": "Assign a name to identify the connection minimum of 5 characters"
```
delliott90, line 8: You may want to put minimum 5 characters in parenthesis otherwise it looks like a run-on sentence.

benjamin-craig: I think it is a good idea.  I added it.

delliott90 (approved):

PR #580, opened 2021-04-05
Enable unmapped_fallback and update mappings

mdazam1942:


delliott90:
Is it possible to add a test for the case where the timestamp is missing from the results, so we can see the mapping fall back do what we expect?

delliott90 (approved):

PR #583, opened 2021-04-07
demo SQL connector

delliott90:


stix_shifter_modules/mysql/requirements.txt
```diff
@@ -0,0 +1 @@
+mysql-connector-python
```
mdazam1942, line 1: its better to set a specific version.

mdazam1942 (approved):

PR #613, opened 2021-05-14
Mapping is no longer hidden

benjamin-craig:
Mapping is no longer hidden

### DIFF
--- a/stix_shifter_modules/config.json
+++ b/stix_shifter_modules/config.json
@@ -60,7 +60,6 @@
             "mapping": {
                 "type": "json",
                 "optional": true,
-                "hidden": false,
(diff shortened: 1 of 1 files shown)


delliott90 (approved):

stix_shifter_modules/config.json
```diff
@@ -60,7 +60,7 @@
             "mapping": {
                 "type": "json",
                 "optional": true,
-                "hidden": true,
+                "hidden": false,
```
yurii-klymenko, line 63: to remove the line would make it a bit cleaner

yurii-klymenko (approved):

PR #617, opened 2021-05-26
Support QRadar offense fields and function

antonpibm:
This PR adds offense related field and function in AQL for QRadar module.
Adding:

- field `hasOffense` - indicate association with and offense
- function `INOFFENSE` - used in where clause only, accepts an offense ID as argument and filters events/flows which relate to that offense. 

Note that `INOFFENSE` doesn't correspond to any field, and is used only in where clause.
Both `hasOffense` and `INOFFENSE` are supported both in events and flows

### DIFF
--- a/stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
+++ b/stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
@@ -71,6 +71,7 @@
     "\"tls ja3s hash\" as tlsja3shash",
     "\"suspect content descriptions\" as suspectcontentdescriptions",
     "flowid",
-    "\"content type\" as contenttype"
+    "\"content type\" as contenttype",
+    "hasOffense"
   ]
 }
--- a/stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
+++ b/stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
@@ -71,7 +71,6 @@
     "\"tls ja3s hash\" as tlsja3shash",
     "\"suspect content descriptions\" as suspectcontentdescriptions",
     "flowid",
-    "\"content type\" as contenttype",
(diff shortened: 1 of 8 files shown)


stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
```diff
@@ -71,6 +71,7 @@
     "\"tls ja3s hash\" as tlsja3shash",
     "\"suspect content descriptions\" as suspectcontentdescriptions",
     "flowid",
-    "\"content type\" as contenttype"
+    "\"content type\" as contenttype",
+    "hasOffense"
```
delliott90, line 75: Would this be different from `hasoffense` which is also in this file?

delliott90:
Do these changes require any new QRadar extensions to be installed? Right now Sysmon and Windows extensions would be required to run the connector with the default mappings.

stix_shifter_modules/qradar/stix_translation/json/aql_flows_fields.json
antonpibm: not at all, missed the older one. The mapping for this field is missing in to_stix and from_stix files, so the only place to remove is here.

antonpibm:
@delliott90 These changes doesn't require any extension and are available in v7.3.3 
https://www.ibm.com/docs/en/qsip/7.3.3?topic=language-event-flow-simarc-fields-aql-queries
https://www.ibm.com/docs/en/qsip/7.3.3?topic=language-aql-data-retrieval-functions#r_aql_supported_functions__INOFFENSE

antonpibm:
@delliott90 @mdazam1942 hey, anything else should be done before this could be merged?

delliott90 (approved):

delliott90:
Looks good, I'll merge it.

PR #622, opened 2021-05-31
PR for Trend Micro Vision One connector

alex-y-lin:


delliott90:
Hi @alex-y-lin, thanks for the submission! Would you be able to sign the CLA? It's a requirement before we would be able to merge this in.

stix_shifter_modules/vision_one/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,31 @@
...
+class ResultsConnector(BaseResultsConnector):
+    def __init__(self, api_client):
+        self.api_client = api_client
+        self.logger = logger.set_logger(__name__)
+
+    def create_results_connection(self, search_id, offset, length):
+        try:
+            min_range = offset
+            if length > 1000:
+                self.logger.warning("The length exceeds length limit. Use default length: 1000")
```
mdazam1942, line 15: is this warning necessary?

stix_shifter_modules/vision_one/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,301 @@
...
+    def _format_like(value) -> str:
+        # value = "'%{value}%'".format(value=value)
+        value = "{value}".format(value=value)
+        return QueryStringPatternTranslator._escape_value(value)
+
+    # pylint: disable=unused-argument
+    @staticmethod
+    def _escape_value(value, comparator=None) -> str:
+        if isinstance(value, str):
+            return '{}'.format(value.replace('\\', '\\\\').replace('\"', '\\"').replace('(', '\\(').replace(')', '\\)'))
```
mdazam1942, line 63: there are lots of unnecessary `\` in the translated query. you may want to update the method with something like this depending on your native query:
`return '{}'.format(value.replace('\"', '\\"').replace('(', '\\(').replace(')', '\\)'))
`

mdazam1942 (changes requested):
Overall most part of the module looks fine. Need to fix few things before we can merge it. Combined Observation Expression doesn't work for some reason. You may want to have a look. here's an example of a valid stix pattern:
`python main.py translate vision_one query "{}" "[ipv4-addr:value = '66.172.27.225'] OR [ipv6-addr:value = '3001:0:0:0:0:0:0:2']"`

One more thing, I think we should add trendmicro as a prefix in the module name. So the module name can be **trendmicro_vision_one** @delliott90 what do you think?

stix_shifter_modules/vision_one/stix_transmission/results_connector.py
alex-y-lin: Users should be informed that if users set length to be greater than 1000, the API will use 1000 instead.

alex-y-lin:
> Overall most part of the module looks fine. Need to fix few things before we can merge it. Combined Observation Expression doesn't work for some reason. You may want to have a look. here's an example of a valid stix pattern and returned error:
> 
> ```
> python main.py translate vision_one query "{}" "[ipv4-addr:value = '66.172.27.225'] OR [ipv6-addr:value = '3001:0:0:0:0:0:0:2']"
> {
>     "success": false,
>     "code": "not_implemented",
>     "error": "wrong parameter : Parenthesis is not support in MDL"
> }
> ```
The dialect, messageActivityData' does not support parenthesis in query.

alex-y-lin:
> Hi @alex-y-lin, thanks for the submission! Would you be able to sign the CLA? It's a requirement before we would be able to merge this in.

I am waiting for my legal team's approval.

stix_shifter_modules/vision_one/entry_point.py
```diff
@@ -0,0 +1,45 @@
...
+            query_connector = QueryConnector(api_client)
+            status_connector = StatusConnector()
+            results_connector = ResultsConnector(api_client)
+            delete_connector = DeleteConnector(api_client)
+
+            self.set_results_connector(results_connector)
+            self.set_status_connector(status_connector)
+            self.set_delete_connector(delete_connector)
+            self.set_query_connector(query_connector)
+            self.set_ping_connector(ping_connector)
```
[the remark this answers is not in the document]
alex-y-lin: setup_transmission_simple uses `connector = module.StatusConnector(api_client)`. Since there is no api_client argument in constructor of Vision One  StatusConnector, the setup logic can not just use setup_transmission_simple.

stix_shifter_modules/vision_one/stix_translation/query_constructor.py
alex-y-lin: Escaping `\` by `\\` is necessary for native query string of Trend Micro Vision One API.

stix_shifter_modules/vision_one/stix_transmission/api_client.py
```diff
@@ -0,0 +1,93 @@
...
+                                    cert_verify=connection.get('selfSignedCert', True),
+                                    sni=connection.get('sni', None)
+                                    )
+
+    def ping_data_source(self):
+        # Pings the data source
+        endpoint = 'v2.0/siem/events'
+        now = datetime.datetime.utcnow().isoformat(timespec="milliseconds") + "Z"
+        resp = self.client.call_api(endpoint, 'GET', timeout=self.timeout, urldata={"startDateTime": now, "endDateTime": now})
+        return {"code": resp.code}
```
mdazam1942, line 39: why are you just returning the response code? PingConnector class calls this method to get the response. response is already handled in `PingConnector.ping_connection()`

stix_shifter_modules/vision_one/stix_transmission/results_connector.py
mdazam1942: so does it mean no matter what the user sets the length of the query it will only return 1000 results?

stix_shifter_modules/vision_one/configuration/config.json
```diff
@@ -0,0 +1,46 @@
...
+        "options": {
+            "timeout": {
+                "type": "number"
+            }
+        }
+    },
+    "configuration": {
+        "auth": {
+            "type": "fields",
+            "tok": {
```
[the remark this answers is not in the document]
mdazam1942: still see `tok`. would you please rename it to  `token`

mdazam1942:
> > Overall most part of the module looks fine. Need to fix few things before we can merge it. Combined Observation Expression doesn't work for some reason. You may want to have a look. here's an example of a valid stix pattern and returned error:
> > ```
> > python main.py translate vision_one query "{}" "[ipv4-addr:value = '66.172.27.225'] OR [ipv6-addr:value = '3001:0:0:0:0:0:0:2']"
> > {
> >     "success": false,
> >     "code": "not_implemented",
> >     "error": "wrong parameter : Parenthesis is not support in MDL"
> > }
> > ```
> 
> The dialect, messageActivityData' does not support parenthesis in query.

So this needs to be handled for STIX Combined Observation Expression. that's a valid query I tried. so if messageActivityData doesn't support parenthesis than you need to handle that in query constructor.

stix_shifter_modules/vision_one/stix_transmission/results_connector.py
alex-y-lin: If the user sets the length less than or equals to 1000, the query will return the amount of data that the user wants. If the user set the length greater than 1000, the query will return 1000 results only.

stix_shifter_modules/vision_one/stix_transmission/api_client.py
alex-y-lin: This API is just to check whether the target server is available and token is valid.

mdazam1942:
@alex-y-lin I think the module name should be `trendmicro_vision_one`. its better to have the vendor name as the module name.

alex-y-lin:
> @alex-y-lin I think the module name should be `trendmicro_vision_one`. its better to have the vendor name as the module name.

The module name is renamed. Would you please review the changes?

stix_shifter_modules/trendmicro_vision_one/stix_transmission/api_client.py
```diff
@@ -0,0 +1,93 @@
...
+        payload = json.loads(search_id)
+        resp_dict = dict()
+        all_data = list()
+        resp_dict["data"] = {"logs": all_data}
+        start = range_start if range_start else 0
+        end = range_end if range_end else 0
+        offset = start
+        fetch_count = 0
+        max_fetch_count = 10
+        while True:
```
mdazam1942, line 56: not sure why you used `while`. try/catch might be more appropriate to gracefully rasise the exceptions.

mdazam1942 (commented):
Please go through the comments and let us know if you have any questions and help to fix some issues.

And also would you please add a readme file for your connector module with some details on the api and kind of data that the connector retrieves with some CLI commands. There are some readme examples in most of the module such as https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/arcsight/stix_translation/README.md

stix_shifter_modules/trendmicro_vision_one/stix_transmission/api_client.py
alex-y-lin: I use for loop instead. The max fetch count for self-protection. If reaching the max fetch count but not retrieve all data yet, just add log for warning and return the fetched data.

stix_shifter_modules/trendmicro_vision_one/configuration/lang_en.json
```diff
@@ -0,0 +1,32 @@
...
+        "port": {
+            "label": "Host Port",
+            "description": "Set the port number that is associated with the Host name or IP"
+        },
+        "help": {
+            "label": "Need additional help?",
+            "description": "More details on the data source setting can be found in the specified link"
+        },
+        "selfSignedCert": {
+            "label": "ArcSight logger Connection Certificate",
```
mdazam1942, line 16: `Trend Micro Connection Certificate`

stix_shifter_modules/trendmicro_vision_one/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,301 @@
...
+
+        elif isinstance(expression, CombinedComparisonExpression):
+            operator = self._lookup_comparison_operator(expression.operator)
+            expression_01 = self._parse_expression(expression.expr1, dialect)
+            expression_02 = self._parse_expression(expression.expr2, dialect)
+            if not expression_01 or not expression_02:
+                return ''
+            if isinstance(expression.expr1, CombinedComparisonExpression):
+                if dialect == 'messageActivityData':
+                    raise NotImplementedError("Parenthesis is not supported in MDL")
```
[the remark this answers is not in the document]
alex-y-lin: I have implemented the limitation of messageActivityData query. The messageActivityData dialect does not support: 1. using AND and OR simultaneously. 2. using multiple criteria for the same field. 3. using NOT.

mdazam1942:
> Please go through the comments and let us know if you have any questions and help to fix some issues.
> 
> And also would you please add a readme file for your connector module with some details on the api and kind of data that the connector retrieves with some CLI commands. There are some readme examples in most of the module such as https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/arcsight/stix_translation/README.md

@alex-y-lin can you please add a readme.md for your module? details in the comment ^^

mdazam1942 (approved):
Thanks for making all the changes. Looks good now.

PR #631, opened 2021-06-17
Added finding attributes for splunk connector 

pramodgslab:


imm007 (changes requested):
Change commit message

PR #633, opened 2021-06-17
Sending validation updates

elephantnine9:


mdazam1942 (approved):
merging as it is. will fix the unittest in a later pr

PR #645, opened 2021-07-12
Msatp fixes

aviv1ron1:
This PR fixes a few erroneus or missing mappings in the MS ATP module.
missing hostname in certain events in `x-oca-asset`
missing IP addresses in `x-oca-asset`
added original message from ms atp as base64 in `x-oca-event.original_ref`
added references to url and process from `x-oca-event`
added mapping to `x-oca-event.action`


aviv1ron1:
Additionaly this merge adds pulling alerts from ms atp.
Uptill today the module was not querying the alerts api - it was returning only raw log events.
Now it will also fetch data from the alerts table.

PR #647, opened 2021-07-21
Move Car mapping to Splunk Module and Rename event to cim

pramodgslab:


mdazam1942 (changes requested):
I think there has been a miscommunications. what we meant was, you can keep the changes that you made in previous pr which is creating a separate mapping for splunk module.

mdazam1942:
@pramodgslab do we still need this PR?

mdazam1942:
closing this since no need for these changes anymore

PR #653, opened 2021-07-26
Onelogin connector

akshay-pange-exa:


stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
```diff
@@ -0,0 +1,61 @@
...
+      "display_name": ["user_name"]
+    }
+  },
+  "ipv4-addr": {
+    "fields": {
+      "value": ["ipaddr"]
+    }
+  },
+
+  "x-onelogin-finding": {
```
delliott90, line 14: Just make this `x-onelogin` to follow the convention of other connector's custom attributes.

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
```diff
@@ -0,0 +1,61 @@
...
+  "x-onelogin-risk": {
+    "fields": {
+      "risk_score": ["risk_score"],
+      "risk_reasons": ["risk_reasons"],
+      "risk_cookie_id": ["risk_cookie_id"],
+      "error_description": ["error_description"]
+    }
+  },
+
+  "x-onelogin-user-account": {
```
delliott90, line 53: There is already an official STIX `user-account` object. Any properties that don't fit with the existing user-account object and be added as custom properties (denoted by an `x_` prefix) on the official STIX object. Ex: `user-account:x_actor_user_id`

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
delliott90: Actually, ignore that suggestion, though I think some of these may fix in the `x-ibm-finding` object, such as policy_name.

stix_shifter_modules/onelogin/configuration/config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "connection": {
+        "type": {
+            "displayName": "onelogin"
```
delliott90, line 4: Should be `OneLogin`.

stix_shifter_modules/onelogin/configuration/lang_en.json
```diff
@@ -0,0 +1,37 @@
...
+        "sni": {
+            "label": "Server Name Indicator",
+            "description": "The Server Name Indicator (SNI) enables a separate hostname to be provided for SSL authentication. Not Required for QRadar on Cloud."
+        }
+    },
+    "configuration": {
+        "auth": {
+            "clientId": {
+                "label": "Client Id",
+                "description": "Client ID of onelogin"
```
delliott90, line 29: Should be "Client ID of OneLogin"

stix_shifter_modules/onelogin/configuration/lang_en.json
```diff
@@ -0,0 +1,37 @@
...
+            },
+            "clientSecret": {
+                "label": "Client Secret",
+                "description": "Client Secret of onelogin"
```
delliott90, line 33: Should be "Client secret of OneLogin"

stix_shifter_modules/onelogin/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,193 @@
...
+                        "proxy_ip": ["ipv4", "ipv4_cidr"],
+                        }
+
+logger = logging.getLogger(__name__)
+
+
+class QueryStringPatternTranslator:
+    # Change comparator values to match with supported data source operators
+    comparator_lookup = {
+        ComparisonExpressionOperators.And: "&",
```
delliott90, line 21: Is this all the supported comparison operations supported by OneLogin? How are OR conditions handled? You'd also need to at least have something for `ObservationOperators.Or` and `ObservationOperators.And`, both of which would equate to an OR statement (or equivalent) on the native data source query. Without this, any pattern with multiple observations would raise an error.

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
```diff
@@ -0,0 +1,61 @@
...
+  "x-onelogin-finding": {
+    "fields": {
+      "client_id": ["client_id"],
+      "directory_id": ["directory_id"],
+      "event_type_id": ["event_type_id"],
+      "unique_id": ["id"],
+      "resolution": ["resolution"],
+      "since": ["since"],
+      "until": ["until"],
+      "proxy_ip": ["proxy_ip"],
```
delliott90, line 23: This should be in a `ipv4-addr` and `ipv6-addr` object. In general, you should be mapping to standard STIX objects when possible. Relying so heavily on custom objects limits the ability of stix-shifter to conduct federated search across multiple data sources.

stix_shifter_modules/onelogin/stix_translation/query_constructor.py
pramodgslab: @delliott90 
Okay. 
OneLogin data source only supports '& (AND)' as a comparison operation.
I think we have to equate '&' for both `ObservationOperators.Or` and `ObservationOperators.And` since our data source (OneLogin) does not supporting OR operator while searching events.  
Like :
        ObservationOperators.And: '&',
        ObservationOperators.Or: '&'

stix_shifter_modules/onelogin/configuration/lang_en.json
pramodgslab: @delliott90 Done

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
pramodgslab: Yes. @delliott90 
Resolved.

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
pramodgslab: Done @delliott90

stix_shifter_modules/onelogin/configuration/config.json
pramodgslab: Done

stix_shifter_modules/onelogin/configuration/lang_en.json
pramodgslab: Done

stix_shifter_modules/onelogin/stix_translation/json/from_stix_map.json
akshay-pange-exa: Yes  @delliott90  moved  `proxy_ip` to `ipv4-addr` . Since OneLogin not support `OR` comparison operator  added as custom field `x_proxy_ip`

mdazam1942 (commented):
@akshay-pange-exa have you looked into using the client SDK? https://developers.onelogin.com/api-docs/2/getting-started/client-sdks

I think you should use the SDK in transmission module which makes the code lot simpler and easy to maintain. For example, to generate the token -

```
from onelogin.api.client import OneLoginClient

client = OneLoginClient( client_id,  client_secret,  region)
token = client.get_access_token()
```

To get event just call:
`events = client.get_events(self, query_parameters=None, max_results=None)`

stix_shifter_modules/onelogin/entry_point.py
```diff
@@ -0,0 +1,30 @@
...
+            status_connector = base_sync_connector
+            results_connector = ResultsConnector(api_client)
+            delete_connector = DeleteConnector(api_client)
+
+            self.set_results_connector(results_connector)
+            self.set_status_connector(status_connector)
+            self.set_delete_connector(delete_connector)
+            self.set_query_connector(query_connector)
+            self.set_ping_connector(ping_connector)
+
```
mdazam1942, line 29: if connection:
            self.setup_transmission_simple(connection, configuration)

stix_shifter_modules/onelogin/stix_transmission/delete_connector.py
```diff
@@ -0,0 +1,9 @@
+from stix_shifter_utils.modules.base.stix_transmission.base_delete_connector import BaseDeleteConnector
+
+
+class DeleteConnector(BaseDeleteConnector):
+    def __init__(self, api_client):
+        self.api_client = api_client
+
+    def delete_query_connection(self, search_id):
+        return {"success": True}
```
mdazam1942, line 9: looks like api doesn't allow deleting the search. so I don't think you need `delete_connector.py` at all.

akshay-pange-exa: Yes @mdazam1942 Removed

stix_shifter_modules/onelogin/entry_point.py
akshay-pange-exa: @mdazam1942 Done

imm007:
> @akshay-pange-exa have you looked into using the client SDK? https://developers.onelogin.com/api-docs/2/getting-started/client-sdks
> 
> I think you should use the SDK in transmission module which makes the code lot simpler and easy to maintain. For example, to generate the token -
> 
> ```
> from onelogin.api.client import OneLoginClient
> 
> client = OneLoginClient( client_id,  client_secret,  region)
> token = client.get_access_token()
> ```
> 
> To get event just call:
> `events = client.get_events(self, query_parameters=None, max_results=None)`

pramodgslab:
@mdazam1942 : 
We were looking into OneLogin client SDK `version 2.0.0`  yesterday. And we were facing issues with that version. There has been upgrade recently and we have upgraded client `version 2.0.1` and now it is working. we'll update code with this version.

pramodgslab:
@mdazam1942 We have used onelogin client SDK (`version 2.0.1`). Please review.

stix_shifter_modules/onelogin/stix_translation/query_constructor.py
delliott90: I think you may need to break up each observation into its own query, so you will return multiple translated queries in the list returned from the query constructor. STIX observations separated by AND must be different observations (ie. events) but with translating the ObservationOperator AND to &, you are looking for one event in OneLogin with both data elements.

akshay-pange-exa: @delliott90 okay we have made changes as per your suggestion please review.

stix_shifter_modules/onelogin/README.md
```diff
@@ -0,0 +1,219 @@
...
+                    "user_id": "138593517",
+                    "x_actor_user_id": 12345,
+                    "x_assuming_acting_user_id": 12345,
+                    "x_actor_user_name": "Mapping",
+                    "display_name": "Firstname Lastname"
+                },
+                "3": {
+                    "type": "ipv4-addr",
+                    "value": "121.0.0.1",
+                    "x_proxy_ip": "121.0.0.1"
```
delliott90, line 130: Does proxy_ip come in from OneLogin as a different value or would it always be the same as `ipaddr`? Since they come in as different fields, I think they should be treated as different STIX `ipv4-addr` objects. It doesn't make sense to have it as a custom property on this object when it may be a different IP, and if it's the same IP then it's just duplicated information.

delliott90 (approved):

PR #672, opened 2021-08-16
Refactor guardium stix-shifter connector to use stix-shifter RestApiClient

raizik:
Fix for SECCON-15552 Removed all requests lib usage


stix_shifter_modules/guardium/stix_transmission/guard_utils.py
```diff
@@ -1,25 +1,29 @@
...
 
-class GuardApiClient(object):
 
-    def __init__ (self,client_id, url, secret, user, password):
-        super().__init__()
+class GuardApiClient(RestApiClient):
+
+    def __init__(self, params, host, port, headers,
+                 url_modifier_function, cert_verify, sni, auth):
+        self.client = RestApiClient(host, port, headers, url_modifier_function, cert_verify, sni, auth)
```
baulus, line 16: I am not quite sure about this. GuardApiClient inherits from RestApiClient and has a property (self.client) that is again RestApiClient...
Is this solution about to use self.client to get token?

raizik: Yes. It uses `self.client` for all of the requests. First - ApiClient and GuardApiClient have a client field and it's legal to have an attribute with the same name in two independent classes. Second - calling `self.client = RestApiClient(host, port, headers, url_modifier_function, cert_verify, sni, auth)` is the same as calling `self.client =  super().__init__(host, port, headers, url_modifier_function, cert_verify, sni, auth)` which is a must in a child class.

Would it be better if I change the name of ApiClient's client field to `aux_client`, or something of the sort, for better readability?

baulus: I guess keeping it as just client is also good in this case

delliott90 (approved):

PR #675, opened 2021-08-24
Secret server UDI Connector

Deepshikha8514:
Secret server UDI Connector


stix_shifter_modules/secretserver/stix_transmission/api_client.py
```diff
@@ -35,11 +36,15 @@ def get_token(self):
...
                                                                             urldata=None,
                                                                             timeout=None)
-        res = response.response.text
-        json_obj = json.loads(res)
-        token = json_obj.get('access_token')
-        self.accessToken = 'Bearer' + " " + token
-        return self.accessToken
+        if(response.code !=200):
+            logging.error("Login failed",response.code )
+            exit(0)
```
mdazam1942, line 41: you can't exit code like this. you need to raise the exception and make sure the error is returned in the error object by calling `ErrorResponder.fill_error()`

mdazam1942 (changes requested):
there are some issues need to be fixed. And also the delete call doesn't work as expected. I suppose the API doesn't have any delete operations so you need to return a hardcoded object in that case. here's an example: https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/mysql/stix_transmission/delete_connector.py

here's a delete command:
 `python main.py transmit secretserver '{"host":"1.1.1.1"}' '{"auth":{"username":"user","password":"pass"}}' delete "eyJxdWVyeSI6ICJTRUxFQ1QgKiBGUk9NIFNlY3JldEV2ZW50RGV0YWlsIFdIRVJFIEV2ZW50U3ViamVjdCBMSUtFICclJSUnIFNUQVJUIHQnMjAxOS0wMS0yOFQxMjoyNDowMS4wMDlaJyBTVE9QIHQnMjAyMS0wNy0xNFQxMjo1NDowMS4wMDlaJyIsICJ0YXJnZXQiIDogImh0dHBzOi8vOS40Ni44Ni4xMjAiLCAicmFuZG9tIjoiMTIifQ=="`

mdazam1942:
I see there's still an account in your name `Deepshikha Sinha` which didn't sign the CLA.

stix_shifter_modules/secretserver/stix_transmission/api_client.py
```diff
@@ -89,7 +99,9 @@ def build_searchId(self):
             self.search_id = s_id
         return s_id
 
-    def get_search_results(self, search_id, index_from=None, fetch_size=None):
+    def get_search_results(self, search_id, index_from=1, fetch_size=2):
+        ind_from = index_from
+        fet_size = fetch_size
```
mdazam1942, line 104: not sure what's the purpose of `ind_from` and `fet_size` and why they are assigned 1 and 2 values respectively.

stix_shifter_modules/secretserver/stix_transmission/delete_connector.py
```diff
@@ -1,24 +1,10 @@
 from stix_shifter_utils.modules.base.stix_transmission.base_delete_connector import BaseDeleteConnector
-from stix_shifter_utils.utils.error_response import ErrorResponder
 from stix_shifter_utils.utils import logger
```
mdazam1942, line 2: you don't use logger here so it can be removed.

PR #685, opened 2021-09-15
fix type property

olegzhr:
because the "type" property need to be inside the "auth" object

### DIFF
--- a/stix_shifter_modules/alertflex/configuration/config.json
+++ b/stix_shifter_modules/alertflex/configuration/config.json
@@ -14,7 +14,7 @@
             "max": 65535
         },
         "help": {
-            "default": "https://github.com/olegzhr/cnode/blob/master/doc/integration.rst",
+            "default": "https://alertflex.github.io/doc/pages/integration.html",
             "type": "link"
         },
         "sni": {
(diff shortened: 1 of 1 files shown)


olegzhr:
This pull request was done according to the recommendation from your colleague Jingqiu Du -

'The issue is we can't see any of the auth fields in the configuration, because the "type" property need to be inside the "auth" object'

After deploy this fix, stix-shifter started response with ERROR:

stix_shifter_utils.utils.error_response ERROR    error occurred: {'missing_params': ['configuration.auth.login'], 'unexpected_params': {'configuration': {'auth': {'username': 'admin'}}}}

stix_shifter_modules/alertflex/configuration/config.json
```diff
@@ -27,14 +27,14 @@
         }
     },
     "configuration": {
-        "type" : "fields",
         "auth": {
-            "username": {
+            "type" : "fields",
+            "login": {
```
delliott90, line 32: If username is changing to login, you will also need to change it here as well https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/alertflex/configuration/lang_en.json#L27.

delliott90 (approved):

PR #708, opened 2021-10-08
SumoLogic UDI Connector

imm007:


imm007:
> > I require one set of `\` for the json query string as the json key-value pairs have to be in double quoted strings and the sumologic search query has to be in double quotes as well.
> > ```
> > {
> >   "query": "(_sourcehost = \"sumologic.gslab.com\")",
> >   "from": "2021-09-20T12:00:00",
> >   "to": "2021-09-26T11:00:00",
> >   "timeZone": "IST",
> >   "byReceiptTime": false
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The other `\` and `\n` are populated when the above json object is converted to string as the translated query has to be in string format.
> > ```
> > "{\"query\": \"(_sourcehost = \\\"sumologic.gslab.com\\\")\",\n\"from\": \"2021-09-20T12:00:00\",\n\"to\": \"2021-09-26T11:00:00\"}"
> > ```
> 
> I don't understand why you need to escape json string. you only need escaping if you need to escape any special characters inside your query string. the standard json quotes and formatting don't need escaping. they are just how python prints json string.

Made changes so multiple backslashes are not visible on query translate

PR #718, opened 2021-10-22
Optimization round 1

baulus:


stix_shifter_utils/stix_translation/src/utils/transformers.py
```diff
@@ -302,6 +302,29 @@ def transform(obj):
             return result
         return obj
 
+
+class CheckIPv6(ValueTransformer):
+    """A value transformer for filtering-out from a list all values which are not valid IPv6 values"""
```
delliott90, line 307: How is this class any different from `FilterIPv6List`? They both have the same description comment. Are you just trying to see if the value is indeed an IPv6?

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,201 @@ def gen_dict_extract(self, key, var):
...
-                    return
-
-            ds_key_def_list = [ds_key_def_obj]
 
 
-        for ds_key_def in ds_key_def_list:
-            if ds_key_def is None or 'key' not in ds_key_def:
-                self.logger.debug('{} is not valid (None, or missing key)'.format(ds_key_def))
-                continue
+    def _check_stix_value_pattern(self, observable_key, return_value):
```
delliott90, line 87: You may want to rename this to `check_stix_value_regex` or something like that. This existing naming could be confusing considering the project translates STIX Patterns, which has nothing to do with this method.

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,201 @@ def gen_dict_extract(self, key, var):
...
+            object_tag_ref_map['tags'][tag] = {'i': tag_ind, 'n': 0}
+            
+        if tag_ind is not None:
+            tag_ind = str(tag_ind)
+            if property_key and '_ref' not in property_key:
+                object_tag_ref_map['non_ref_props'][tag_ind] = True
+
+        return tag_ind
+
+    def _add_prperty(self, type_name, property_key, parent_key_ind, value, objects, group=False, cybox=True):
```
mdazam1942, line 157: just a typo fix _add_prperty -> _add_property

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,201 @@ def gen_dict_extract(self, key, var):
...
+                objects[parent_key_ind][property_key] = value
+            elif isinstance(value, dict):
+                objects[parent_key_ind][property_key] = dict_merge(objects[parent_key_ind][property_key], value)
+            elif isinstance(objects[parent_key_ind][property_key], list) and group:
+                objects[parent_key_ind][property_key].extend(value)
+            else: #TODO: get rid of this weird 'else'
+                # print({'old': objects[parent_key_ind][property_key], 'new': value})
+                pass
+
+    def _process_properties(self, to_stix_config_prop, data, objects, object_tag_ref_map, parent_data=None, ds_sub_key=None, object_key_ind=None):
```
mdazam1942, line 179: you may wanna avoid using names like process. it may confuse people because process is a name of a stix object. I thought we you are just working on process properties :)

stix_shifter_utils/stix_translation/src/utils/transformers.py
baulus: Will change the description. This does the same as `FilterIPv6List` but for one element. If the value is not IPv6, the value will be transformed to null.

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
baulus: I can use the `_valid_stix_value` as its ancestor. 
https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py#L151
the meaning of the both functions is pretty my the same.

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
baulus: agree, just didn't come up with more proper name. 
this function is kind of similar to the previous `_tranform` [here](https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py#L209), 
but for some reason I did not like it, because there is "public" function `transform` already.

stix_shifter_utils/utils/helpers.py
```diff
@@ -0,0 +1,43 @@
+def dict_merge(dct, merge_dct, add_keys=True):
```
baulus, line 1: I pretty much used from here, just enhanced with list merge:
https://gist.github.com/angstwad/bf22d1822c38a92ec0a9?permalink_comment_id=2622319#gistcomment-2622319

mdazam1942: I would put that link in the comment for reference

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -1,7 +1,9 @@
 import re
+from traceback import print_exception
```
mdazam1942, line 2: I think its unused import. you can remove it if you don't need it

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,225 @@ def gen_dict_extract(self, key, var):
...
+        :rtype: bool
+        """
+        try: 
+            if observable_key in self.properties and 'valid_regex' in self.properties[observable_key]:
+                pattern = re.compile(self.properties[observable_key]['valid_regex'])
+                match = pattern.match
+                if not match(str(stix_value)):
+                    return False
+            return True
+        except Exception as e:
```
mdazam1942, line 102: its better to print some debug log when this Exception hits

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,225 @@ def gen_dict_extract(self, key, var):
...
-                            for obj_name in object_map:
-                                if pattern.match(obj_name):
-                                    val = object_map.get(obj_name)
-                                    stix_value.append(val)
-                        else:
-                            stix_value = object_map.get(references)
-                            if not self._valid_stix_value(self.properties, key_to_add, stix_value):
-                                continue
+                        return_value = self._get_tag_ind(references, object_tag_ref_map, create_on_absence=False)
+                        if unwrap is not False and not isinstance(return_value, list):
```
mdazam1942, line 135: there used to be a comment about `unwrap` flag. I would keep the comment here or somewhere else whatever is suitable.

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -190,169 +82,225 @@ def gen_dict_extract(self, key, var):
...
+                property_key: value
+            }
+        else:
+            if not property_key in objects[parent_key_ind]:
+                objects[parent_key_ind][property_key] = value
+            elif isinstance(value, dict):
+                objects[parent_key_ind][property_key] = dict_merge(objects[parent_key_ind][property_key], value)
+            elif isinstance(objects[parent_key_ind][property_key], list) and group:
+                objects[parent_key_ind][property_key].extend(value)
+            else: #TODO: get rid of this weird 'else'
```
mdazam1942, line 192: just wondering if we should keep this else condition

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
mdazam1942: then how would we know if this validity check fails. is raising exception less costly?

mdazam1942 (approved):

stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
baulus: as we discussed, I will add the debug because it is necessary.

delliott90 (approved):

PR #730, opened 2021-11-08
Add stix-shifter CLI parameters to configure max returned results and saving to a file

crahan:
This PR adds `-r`/`--results` and `-f`/`--file` commandline parameter to the `execute` command to customize the maximum number of query results to be printed (with a default value of 10) and save the STIX query result to a file.

For example to print a maximum of 50 results and save the output to `output.json`, use the following CLI command:
```
stix-shifter execute <TRANSMISSION MODULE NAME> <TRANSLATION MODULE NAME> '<STIX IDENTITY OBJECT>' '<CONNECTION OBJECT>' '<CONFIGURATION OBJECT>' '<STIX PATTERN>' --results 50 --file output.json
```

This helps to debug full search results.


mdazam1942 (commented):
Thanks for the improvements. Would you please update the documentation as well? https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/OVERVIEW.md#execute

yurii-klymenko:
instead of doing 
`stix-shifter execute <TRANSMISSION MODULE NAME> <TRANSLATION MODULE NAME> '<STIX IDENTITY OBJECT>' '<CONNECTION OBJECT>' '<CONFIGURATION OBJECT>' '<STIX PATTERN>' --results 50 --file output.json`
you have to call it like:
`stix-shifter execute <TRANSMISSION MODULE NAME> <TRANSLATION MODULE NAME> '<STIX IDENTITY OBJECT>' '<CONNECTION OBJECT>' '<CONFIGURATION OBJECT>' '<STIX PATTERN>' --results 50 > output.json`
no extra param needed.

crahan:
@yurii-klymenko I've tested that and because the output is not using `print` statements, but logged instead, a shell redirect fails. Even if it works, a shell redirect would also output all other info logging entries alongside the actual returned JSON structured data.

crahan:
@mdazam1942 I've updated the PR with the requested documentation update. I've added 2 entries (similar to the section for the `--debug` option) that explain the additional `--file` and `--results` options.

yurii-klymenko:
check out command from my PR https://github.com/opencybersecurityalliance/stix-shifter/pull/735
debug output goes to the console, stix goes to the file.

crahan:
@yurii-klymenko the redirect works for the `transmit` command because the output from that command uses `print()` (versus `log.info()` for the `execute` command).

Here's the Python code for the `transmit` command:
```
elif args.command == TRANSMIT:
    result = transmit(args)  # stix_transmission

print(json.dumps(result, indent=4, sort_keys=False))
```

The `execute` command however uses the following Python code to output the data:
```
log.info('STIX Results: \n' + json.dumps(result, indent=4, sort_keys=False))
```

Using `print()` sends the information to stdout and will allow you to redirect the output (as you demonstrated in your example) and thus a shell redirect will work as expected. The output from the `transmit` command also outputs pure JSON and no additional logging information.

In contrast, the output from the `execute` command has both timestamped logging information (orange in the screenshot) and STIX result data (green in the screenshot).

Screen Shot 2021-11-11 at 01 18 29

The only alternative solution that would work is to not use `log.info()` to output the STIX result, but to use a `print()` statement for the `execute` command. The problem there however is that it wouldn't be clear to the user what output is coming from the logger and what is coming from any `print()` statements.

I'll let @mdazam1942 and @delliott90 chime in with their preferred option:
1. keep `log.info()` to output the STIX result as a log entry and add a `--file` parameter to alternatively save the STIX result to a file.
2. change the `log.info()` line for the `execute` command to a `print()` statement and remove the `--file` parameter. The `print()` statement would then output to stdout and allow a shell redirect to redirect just the STIX result to a file (keeping the remaining log messages to output to the terminal).

mdazam1942:
@crahan as this is CLI options so I think we should follow @yurii-klymenko's suggestion to redirect the stdout to save the bundle or return objects to a file. We can still use your solution to specify the `-r/--results` option. 

PR #735 is merged. So the stdout issue is fixed. Thomas, can you please resolve the conflict and update your PR.

crahan:
@mdazam1942 thanks for the feedback. I've removed the `--file` option and split the log INFO statement to a log message + printed JSON output which can now be captured with a stdout shell redirect.

mdazam1942 (changes requested):
@crahan thanks for the update. would you please resolve the conflicts?

crahan:
My apologies, it looks like updating from develop to get the other PR changes pulled in some additional files
.

mdazam1942 (approved):

PR #738, opened 2021-11-17
update readme to specify supported python development versions and some links update

mdazam1942:


adapter-guide/develop-stix-adapter.md
```diff
@@ -24,7 +24,7 @@ By implementing a new connector, _Developer A_ allows _Product A_ to fit into th
 
 ## Prerequisites
 
-- Your development environment must use Python 3.6.
+- Your development environment must use any of the follow Python versions: 3.6, 3.7, 3.8 and 3.9.
```
delliott90, line 27: Even though 3.6 works, I wonder if we should remove it since it's now deprecated.

mdazam1942: I don't think its officially deprecated by python yet. developers can still use it. https://www.python.org/dev/peps/pep-0494/
we can keep the official support for few more months may be until early next year.

PR #744, opened 2021-12-01
handling unmapped operators in stix pattern

mdazam1942:


stix_shifter_modules/azure_sentinel/configuration/operators.json
```diff
@@ -0,0 +1,17 @@
+{
```
delliott90, line 1: Did you mean to have this file here in the configuration folder? It contains different (and extra) mappings from the operators.json included in the json folder.

mdazam1942: ohh this is a mistake. will remove it

stix_shifter_modules/infoblox/stix_translation/query_constructor.py
```diff
@@ -65,42 +65,14 @@ class QueryStringPatternTranslator:
...
         qualified_queries (list): List of Infoblox translated & formatted queries. Queries are simply the HTTP query parameters for the specific APIs.
     """
-    comparator_lookup = {
-        'tideDbData': {
-            ComparisonExpressionOperators.And: "&",
-            ObservationOperators.And: '&',
-            ComparisonExpressionOperators.Or: "&",
-            ObservationOperators.Or: '&',
-            ComparisonComparators.Equal: "=",
-            ComparisonComparators.GreaterThan: "=",
```
delliott90, line 80: Infoblox seems to be the only reason you need to allow for dialect-specific operator files. But looking at this original mapping, this seems to be wrong and misleading. GreaterThan, LessThan, and so on are not the same as equals and would be misleading if we treat them as such in the translated query. Would it not have been better to just keep them as unsupported operators or am I missing something? I know you didn't develop this connector, this just jumped out at me is all. I think this connector should have just used the one set of operators and have it a smaller set at that. It also seems like the data source doesn't support any sort of OR operator.

stix_shifter_utils/stix_translation/src/utils/unmapped_attribute_stripper.py
```diff
@@ -1,13 +1,15 @@
 from stix_shifter_utils.stix_translation.src.patterns.pattern_objects import StartStopQualifier, ObservationExpression, \
-    ComparisonExpression, Pattern, CombinedComparisonExpression, CombinedObservationExpression
+    ComparisonExpression, Pattern, CombinedComparisonExpression, CombinedObservationExpression, \
+    ComparisonComparators,  ComparisonExpressionOperators, ObservationOperators
```
delliott90, line 3: Are these imports used?

mdazam1942: not really. I probably imported them for some testing and forgot to remove

stix_shifter_modules/qradar/tests/stix_translation/qradar_stix_to_aql/test_qradar_events_stix_to_query.py
```diff
@@ -439,3 +439,17 @@ def test_inoffense_query(self):
...
+        where_statement = "WHERE (INCIDR('127.0.0.1/24',sourceip) OR INCIDR('127.0.0.1/24',destinationip) OR INCIDR('127.0.0.1/24',identityip)) {} {}".format(default_limit, default_time)
+        _test_query_assertions(query, selections, from_statement, where_statement)
+    
+    def test_unmapped_operators_error(self):
+        stix_pattern = "[ipv4-addr:value ISSUPERSET '127.0.0.1/24']"
+        query = _translate_query(stix_pattern)
+        assert query['success'] == False
+        assert ErrorCode.TRANSLATION_MAPPING_ERROR.value == query['code']
+        assert "data mapping error : Unable to map the following STIX objects and properties: " \
+            "['ipv4-addr:value'] or Operators: [IsSuperSet] to data source fields" in query['error']
```
delliott90, line 455: Showing both of these kind of masks what is actually causing the error. I know in this case that it's the operator, but it might be good to split the errors up, or put some conditions in the error code to choose what gets printed out.

mdazam1942: yeah I thought about it. But my intention was showing the users not only the unmapped operators but also the objects that are associated with the error. if this is not the correct approach then sure I can separate the errors.

delliott90: I see your point. You want to show the context where that operator is used. Maybe have three conditions for the error message: 1. If you have both unmapped fields and operators, you show what you have here, just replace the wording to read "and" Operators. 2. If you have just unmapped fields, remove any mention of operators. 3. If you have just unmapped operators, remove mention of unmapped objects and properties, but still show the errored query.

mdazam1942: updated the error messages

stix_shifter/stix_translation/stix_translation.py
```diff
@@ -16,6 +16,8 @@
 DIALECTS = 'dialects'
 SUPPORTED_ATTRIBUTES = "supported_attributes"
 MAPPING_ERROR = "Unable to map the following STIX objects and properties:"
```
delliott90, line 18: I think you need to now add "and operators" to this error message.

mdazam1942: I actually have added that later where exception thrown in line 110-
```
raise DataMappingException(
                                 "{} {} and Operators: {} to data source fields".format(MAPPING_ERROR, unmapped_stix_collection, unmapped_operator_collection)
                             )
```

delliott90 (approved):

PR #748, opened 2021-12-08
testing permissions

sparrell:
this is testing permissions - don't expect it to be accepted (but it can test the process of rejecting)


delliott90:
We had a governance document in progress but [closed the PR](https://github.com/opencybersecurityalliance/stix-shifter/pull/420) since the OCA already has a [governance document](https://github.com/opencybersecurityalliance/oasis-open-project/blob/master/GOVERNANCE.md). Maybe we should have one specific to this project though. @Roseann-Guttierrez we had talked about this before. It would be no problem to add one.

delliott90:
I have reopened the governance PR though I think we will need to make some edits before merging it in. I'm going to close this pull request since that should address the questions brought up in your changes to the contributing file.

PR #813, opened 2022-02-17
results metadata support

yurii-k7:


stix_shifter_modules/stix_bundle/stix_transmission/connector.py
```diff
@@ -61,7 +76,16 @@ def ping_connection(self):
...
-    def create_results_connection(self, search_id, offset, length):
+    def create_results_connection(self, search_id, offset, length, metadata=None):
+
+        #METADATA_SAMPLE metadata_result_id = 0
+        #METADATA_SAMPLE if metadata:
+        #METADATA_SAMPLE     metadata_result_id = int(metadata.get('metadata_result_id', 0))
+        #METADATA_SAMPLE else:
+        #METADATA_SAMPLE     metadata = {}
+        #METADATA_SAMPLE metadata_result_id += 1
+        #METADATA_SAMPLE metadata['metadata_result_id'] = metadata_result_id
```
mdazam1942, line 87: Trying to understand this usage-
1. what is the use of `metadata_result_id`?
2. why does it have to be integer?
3. why are you incrementing the id?
4. what if there's a different key other than `metadata_result_id`?

yurii-klymenko: it is just a sample, any data can be returned in metadata status. That returned metadata will be provided to future status and results calls.

stix_shifter_utils/modules/base/stix_transmission/base_connector.py
```diff
@@ -89,14 +89,23 @@ def ping_connection(self):
...
-        result = entry_point.create_results_connection(search_id, offset, length)
+        if metadata:
+            result = entry_point.create_results_connection(search_id, offset, length, metadata)
+        else:
+            result = entry_point.create_results_connection(search_id, offset, length)
         stats.append({'action': 'transmission', 'time': int(time.time()*1000)})
+        metadata = None
+        if 'metadata' in result:            
+            metadata = result['metadata']
+            del result['metadata']
```
mdazam1942, line 102: just wondering, is it just to avoid passing metadata in results translation?

yurii-klymenko: yes, no need for metadata in translation.

stix_shifter_modules/stix_bundle/stix_transmission/connector.py
mdazam1942: I think, this example would confuse people. Also, I would put example inside dummy connectors instead of stix-bundle `stix_shifter_modules/synchronous_dummy` / `stix_shifter_modules/async_dummy`

delliott90 (approved):

mdazam1942 (approved):

PR #836, opened 2022-03-04
Secret server 1.9

Deepshikha8514:
Name change of secret server to IBM Security Verify Privilege


stix_shifter_modules/secretserver/stix_transmission/api_client.py
```diff
@@ -112,11 +112,9 @@ def get_search_results(self, search_id, index_from, fetch_size):
             else:
                 self.startDate = date.today()
                 self.endDate = self.startDate - timedelta(days = 1)
-            resp = self.get_response()
-            page_size =100
-            resp=resp[(index_from * page_size):(fetch_size * page_size)]
-            return resp
-
+            response = self.get_response()
```
delliott90, line 115: Did you mean to remove the paging when fetching the results?

Deepshikha8514: yes, We do not need pagination in our implementation. As we are getting whole result in one response (page). and api which we are using for fetching the result that does not support pagination.

delliott90 (approved):

PR #840, opened 2022-03-07
Add IN operator into onelogin connector

JingqiuDu:


stix_shifter_modules/onelogin/test/stix_translation/test_onelogin_stix_to_query.py
```diff
@@ -227,6 +227,12 @@ def test_risk_cookie_id_query(self):
         queries = 'risk_cookie_id=5&limit=1000'
         _test_query_assertions(query, queries)
 
+    def test_in_operator(self):
+        stix_pattern = "[x-onelogin-finding:operation_name IN ('Default', 'notDefault') ]"
+        query = translation.translate('onelogin', 'query', '{}', stix_pattern)
+        queries = 'operation_name=Default or operation_name=notDefault &limit=1000'
```
mdazam1942, line 233: does this translated query works? I see few whitespaces in the query string which might fail the api query. also you need to confirm whether this ` or `  operator supports by the api.

stix_shifter_modules/onelogin/stix_translation/json/operators.json
```diff
@@ -1,6 +1,7 @@
 {
     "ComparisonExpressionOperators.And": "&",
     "ComparisonComparators.Equal": "=",
+    "ComparisonComparators.In": "or",
```
delliott90, line 4: I don't think this is going to work. The native Onelogin query doesn't support OR. The work around for supporting the IN operator depends on the connector already having defined `ComparisonExpressionOperators.Or` in this object, which it doesn't.

JingqiuDu: So we don't do the IN operator anymore?

PR #845, opened 2022-03-10
Add IN operator for sumologic connector

JingqiuDu:


stix_shifter_modules/sumologic/tests/stix_translation/test_sumologic_stix_to_query.py
```diff
@@ -141,6 +141,14 @@ def test_user_account_created_query(self):
                   "\"toTime\": \"%s\"}" % (from_time, to_time)
         _test_query_assertions(query, queries)
 
+    def test_in_operator(self):
+        stix_pattern = "[user-account:display_name IN ('abc', 'def')]"
+        query = translation.translate('sumologic', 'query', '{}', stix_pattern)
+        _, from_time, to_time = query_constructor.convert_timestamp(query)
+        queries = "{\"query\": \"displayName = (abc OR def)\", \"fromTime\": \"%s\", \"toTime\": \"%s\"}" \
```
mdazam1942, line 148: are you sure this is a valid translated query? have you checked the API if it has any IN operator or similar kind of operator support?  if there's no IN operator support exists then it should work like OR operator.

JingqiuDu: Hi Azam, 

I refer to the sumologic query doc here https://help.sumologic.com/05Search/Get-Started-with-Search/How-to-Build-a-Search/Keyword-Search-Expressions#examples

IN operator is NOT supported naturally from sumologic query, we will need to translate it to an combination of "=" and "OR", there are some examples in the docs using "=" and "OR"

stix_shifter_modules/sumologic/tests/stix_translation/test_sumologic_stix_to_query.py
```diff
@@ -141,6 +141,14 @@ def test_user_account_created_query(self):
                   "\"toTime\": \"%s\"}" % (from_time, to_time)
         _test_query_assertions(query, queries)
 
+    def test_in_operator(self):
+        stix_pattern = "[user-account:display_name IN ('abc', 'def')]"
+        query = translation.translate('sumologic', 'query', '{}', stix_pattern)
+        _, from_time, to_time = query_constructor.convert_timestamp(query)
+        queries = "{\"query\": \"displayName = abc OR displayName = def\", \"fromTime\": \"%s\", \"toTime\": \"%s\"}" \
```
mdazam1942, line 148: the value type should be string. `displayName = "abc"`

stix_shifter_modules/sumologic/stix_translation/query_constructor.py
```diff
@@ -148,6 +155,11 @@ def _parse_expression(self, expression, qualifier=None) -> str:
 
             comparison_string = self._parse_mapped_fields(self, expression, value, comparator, stix_field,
                                                           mapped_fields_array)
+
+            # IN operator logic
+            if expression.comparator == ComparisonComparators.In:
+                comparison_string = self._format_in(mapped_fields_array, expression.value)
```
mdazam1942, line 161: why do you need generate `comparison_string` twice?  is line 156/157 necessary if parsing IN operator? may be implement IN support inside `self._parse_mapped_fields()` or add a check so that you don't need to call self._parse_mapped_fields()

mdazam1942 (approved):

PR #893, opened 2022-04-11
IBM Security Verify connector fixes and additional mapping with verify fields

kr-riteshsinha:
1. Additional validation for multiple filter criteria
2. delete, geo-location  fields are not searchable.
3. 'add' field is added in mapping file.
4. fixed map_attribute issue which is specific to 1.9 version


delliott90:
Why is this PR showing every file as new? I should only be seeing the files that have been changed. Also, there are a lot of conflicts that need to be resolved.

kr-riteshsinha:
I will raise new pull request once the branch is rebase and stable.

PR #912, opened 2022-04-28
stix2matcher.matcher usage optimization

baulus:


stix_shifter_modules/error_test/stix_transmission/connector.py
```diff
@@ -33,19 +32,9 @@ def __init__(self, connection, configuration):
...
-        if found_bindings:
-            matching_sdos = []
-            for binding in found_bindings:
-                matches = [match for match in matcher.get_sdos_from_binding(binding) if match not in matching_sdos]
-                matching_sdos.extend(matches)
-        else:
-            matching_sdos = []
-
+        compiled_pattern = Pattern(pattern)
+        matching_sdos = compiled_pattern.match(observed_data_sdos, verbose)
```
mdazam1942, line 36: so the match() doesn't return just the first anymore?

stix_shifter_modules/stix_bundle/stix_transmission/connector.py
```diff
@@ -31,19 +30,7 @@ def __init__(self, connection, configuration):
...
-
-        found_bindings = matcher.matched()
-
-        if found_bindings:
-            matching_sdos = []
-            for binding in found_bindings:
-                matches = [match for match in matcher.get_sdos_from_binding(binding) if match not in matching_sdos]
-                matching_sdos.extend(matches)
-        else:
-            matching_sdos = []
```
mdazam1942, line 45: so the match() doesn't return just the first anymore?

mdazam1942: looks like this optimization introduced the original problem where the match() function doesn't return all the matching observables.

baulus: there was such problem before?

PR #948, opened 2022-05-18
Added matcher lib support for 2.1

baulus:


requirements.txt
```diff
@@ -0,0 +1,19 @@
+adal==1.2.7
```
delliott90, line 1: top-level requirements shouldn't be checked in since it's generated by the generate_requirements.py script.

PR #964, opened 2022-05-26
Reaqta Connector:Update mapping and unittest

mdazam1942:


stix_shifter_modules/reaqta/stix_translation/json/stix_2_1/to_stix_map.json
```diff
@@ -1379,7 +1379,7 @@
       },
       "etwWorkstation": {
         "key": "x-oca-asset.host_id",
-        "object": "etw_asset"
+        "object": "asset"
```
pcoccoli, line 1382: Does this need to match etwWorkstationName below?  That also uses `etw_asset`.

stix_shifter_modules/reaqta/test/stix_translation/test_reaqta_json_to_stix.py
```diff
@@ -496,4 +496,10 @@ def test_cybox_observables_network_traffic_inbound(self):
...
         assert(extensions['address_family'] == 'IPv4')
-        assert(extensions['outbound'] == False)
\ No newline at end of file
+        assert(extensions['outbound'] == False)
+
+        x_oca_asset = TestReaqtaResultsToStix.get_first_of_type(objects.values(), 'x-oca-asset')
+        ip_refs =  x_oca_asset['ip_refs']
+        obj_num = ip_refs[0]
+        ip_obj = objects[obj_num]
+        assert(ip_obj['value'] == DATA_REMOTE_IP)
```
pcoccoli, line 505: Should the asset IP be the local IP?  The connection is this test is inbound, meaning the dest/server/responder is local (i.e. the asset) and the source/client/originator is remote.

mdazam1942: correct. so the global variable `DATA_REMOTE_IP` is declared by taking the original value from the test results set.  the results translator switch the remote to local ip for inbound conn which is actually DATA_REMOTE_IP -> local ip. I can put a comment there for clarification.

stix_shifter_modules/reaqta/stix_translation/json/stix_2_1/to_stix_map.json
mdazam1942: ohh that's a mistake. I should have kept the `etw_asset` object name. reverting back

pcoccoli (approved):

PR #977, opened 2022-06-02
Added reaqta from_stix generate script

baulus:


mdazam1942 (commented):
we can't have these script files inside the module. it will just add to the size of the image unless you put a logic to skip this folder in setup.py

baulus:
> 
SKIP.ME did not helped, so I included "tools" folder in exclude list

PR #996, opened 2022-06-17
fixed last and first observed and rename label for configuration

kr-riteshsinha:
1. added last observed and first observed attribute.
2. Change host name to tenant name in lang_en.json file
3. removed port parameter for datasource configuration.


stix_shifter_modules/ibm_security_verify/stix_translation/json/to_stix_map.json
```diff
@@ -116,6 +116,30 @@
       "key": "x-oca-event.created",
       "transformer": "EpochToTimestamp",
       "object": "ocaevent"
+    },
+    {
```
mdazam1942, line 120: why do you need `time` field mapped to first+last observed when you already mapped them to startime and endtime?

PR #998, opened 2022-06-20
Replace  all password type fields to certificate type for consistency

mdazam1942:


stix_shifter_modules/alertflex/configuration/config.json
```diff
@@ -22,18 +22,18 @@
...
-            "type": "password",
+            "type": "certificate",
             "optional": true
         }
     },
     "configuration": {
         "auth": {
             "type" : "fields",
             "username": {
-                "type": "password"
```
delliott90, line 33: This should not even be hidden as a password but be plain text I would think.

delliott90:
We don't want all passwords changed over, just the actual certificate fields.

mdazam1942:
> We don't want all passwords changed over, just the actual certificate fields.

ohk the issue description mislead me. will have to revert back all these change.

mdazam1942:
closing this pr. reverting back is more work than just creating new PR  #1000

PR #1005, opened 2022-06-27
Fixed unique_cybox_objects storing

baulus:


pcoccoli (commented):
The change looks correct to me, but it appears that the unit tests have hard-coded IDs that need to be updated.

pcoccoli (approved):
Looks OK to me.

PR #1025, opened 2022-07-05
Removed allowed methods from Retry (passed by default)

snaqati:
allowed methods parameter is causing an issue for the Infoblox module.
The methods passed are default allowed methods specified in the Retry module.

### DIFF
--- a/stix_shifter_utils/stix_transmission/utils/RestApiClient.py
+++ b/stix_shifter_utils/stix_transmission/utils/RestApiClient.py
@@ -103,8 +103,7 @@ class RestApiClient:
                 url = 'https://' + self.server_ip + '/' + endpoint
             try:
                 session = requests.Session()
-                retry_strategy = Retry(total=self.retry_max, backoff_factor=0, status_forcelist=[429, 500, 502, 503, 504],
-                                       allowed_methods=["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"])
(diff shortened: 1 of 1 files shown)


stix_shifter_utils/stix_transmission/utils/RestApiClient.py
```diff
@@ -103,8 +103,7 @@ def call_api(self, endpoint, method, headers=None, data=None, urldata=None, time
                 url = 'https://' + self.server_ip + '/' + endpoint
             try:
                 session = requests.Session()
-                retry_strategy = Retry(total=self.retry_max, backoff_factor=0, status_forcelist=[429, 500, 502, 503, 504],
-                                       allowed_methods=["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"])
+                retry_strategy = Retry(total=self.retry_max, backoff_factor=0, status_forcelist=[429, 500, 502, 503, 504])
```
mdazam1942, line 106: what is the error when allowed_methods is specified? Is there any specific method that infoblox need?

mdazam1942: what version of `urllib3` and `requests` are you using? seems like you may need to upgrade these specified packages in your python env.

snaqati: I'll close the PR.
Thanks again

snaqati:
The issue was caused due to outdated python packages

PR #1064, opened 2022-08-04
CrowdStrike: Added User-Agent string to API Client for tracking

PacketJockey:
Added User-Agent string to API Client for tracking

### DIFF
--- a/stix_shifter_modules/crowdstrike/stix_transmission/api_client.py
+++ b/stix_shifter_modules/crowdstrike/stix_transmission/api_client.py
@@ -43,7 +43,7 @@ class APIClient:
         data = dict()
         headers['Content-Type'] = 'application/json'
         headers['Accept'] = 'application/json'
-        headers['user-agent'] = 'ibm_stixshifter_1.0'
+        headers['user-agent'] = 'oca_stixshifter_1.0'
         headers['Authorization'] = f'Bearer {self.get_token()}'
         endpoint = self.INCIDENTS_IDS_ENDPOINT
         data['filter'] = filter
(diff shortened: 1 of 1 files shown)


stix_shifter_modules/crowdstrike/stix_transmission/api_client.py
```diff
@@ -43,6 +43,7 @@ def get_detections_IDs(self, filter, limit, sort=None):
         data = dict()
         headers['Content-Type'] = 'application/json'
         headers['Accept'] = 'application/json'
+        headers['user-agent'] = 'ibm_stixshifter_1.0'
```
mdazam1942, line 46: I think the user-agent needs to be more generic not ibm specific since connector can be used by non ibm products. let us know if you have any specific reason to set that name.

PacketJockey: How about oca_stixshifter_1.0 ?

mdazam1942: does 1.0 mean anything? if not then i would remove that.

PacketJockey: Yes, it is part of our API standards at CRWD.  As we contribute to the CrowdStrike module we will update it.  It helps in the long term when we are troubleshooting issues that way we know which clone they are working from.

mdazam1942: Sounds good.

mdazam1942 (approved):

PR #1071, opened 2022-08-09
Msatp latest

AmitHaim1:
Add support for x-ibm-finding, x-ibm-ttp-tagging.
Add the ability to query directly from Alerts table, according to basic fields, and then query the appropriate events in
order to extend the results.


mdazam1942:
@AmitHaim1 PR branch is outdated, you need to update your fork and then create a new PR with your changes

PR #1075, opened 2022-08-11
Adding GCP Chronicle UDI Connector

lakshmi51974368:
Adding GCP Chronicle UDI Connector


stix_shifter_modules/gcp_chronicle/configuration/lang_en.json
```diff
@@ -0,0 +1,24 @@
...
+        "host": {
+            "label": "Management IP address or Hostname",
+            "description": "Specify the IP address or  hostname of the data source so that IBM Cloud Pak for Security can communicate with it"
+        },
+        "help": {
+            "label": "Need additional help?",
+            "description": "More details on the data source setting can be found in the specified link"
+        },
+         "selfSignedCert": {
+                "label": "Private Key - Mandatory Field",
```
mdazam1942, line 12: Private Key (Required)

stix_shifter_modules/gcp_chronicle/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,579 @@
...
+
+    @staticmethod
+    def load_json(rel_path_of_file):
+        """
+        Consumes a json file and returns a dictionary
+        :param rel_path_of_file: str
+        :return: dict
+        """
+
+        _json_path = path.dirname(path.realpath(__file__)) + "/" + rel_path_of_file
```
mdazam1942, line 51: read from absolute path.

stix_shifter_modules/gcp_chronicle/configuration/lang_en.json
lakshmi51974368: Updated the label in lang_en.json as suggested

stix_shifter_modules/gcp_chronicle/stix_translation/query_constructor.py
lakshmi51974368: Updated realpath to abspath as suggested.

stix_shifter_modules/gcp_chronicle/stix_transmission/ping_connector.py
```diff
@@ -0,0 +1,59 @@
...
+            ErrorResponder.fill_error(return_obj, response_dict, ['message'], connector=self.connector)
+
+        except Exception as ex:
+            if "timed out" in str(ex):
+                response_dict['code'] = 120
+                response_dict['message'] = str(ex)
+            else:
+                response_dict['message'] = ex
+            self.logger.error('error when getting search results: %s', ex)
+            ErrorResponder.fill_error(return_obj, response_dict, ['message'], connector=self.connector)
```
mdazam1942, line 57: invalid private key returns unknown error code.  assign it to a proper error code
```
{
    "success": false,
    "connector": "gcp_chronicle",
    "error": "gcp_chronicle connector error => ('Could not deserialize key data. The data may be in an incorrect format, it may be encrypted with an unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit parameters).', [_OpenSSLErrorWithText(code=503841036, lib=60, reason=524556, reason_text=b'error:1E08010C:DECODER routines::unsupported')])",
    "code": "unknown"
}
```

mdazam1942 (changes requested):

stix_shifter_modules/gcp_chronicle/configuration/lang_en.json
```diff
@@ -0,0 +1,24 @@
...
+            "label": "Management IP address or Hostname",
+            "description": "Specify the IP address or  hostname of the data source so that IBM Cloud Pak for Security can communicate with it"
+        },
+        "help": {
+            "label": "Need additional help?",
+            "description": "More details on the data source setting can be found in the specified link"
+        },
+         "selfSignedCert": {
+                "label": "Private Key (Required)",
+                "description": "Private Key which is mandatory field used in authentication to make API calls"
```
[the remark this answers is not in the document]
mdazam1942: please update the description to above comment

stix_shifter_modules/gcp_chronicle/stix_translation/transformers.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def transform(obj):
+        pattern = re.compile(
+            r"^([a-zA-Z][a-zA-Z0-9+.-]*):(?:\/\/((?:(?=((?:[a-zA-Z0-9-._~!$&'()*+,;=:]|%[0-9a-fA-F]{2})*))(\3)@)?(?=("
+            r"(?:\[?(?:::[a-fA-F0-9]+(?::[a-fA-F0-9]+)?|(?:[a-fA-F0-9]+:)+(?::[a-fA-F0-9]+)+|(?:[a-fA-F0-9]+:)+(?::|("
+            r"?:[a-fA-F0-9]+:?)*))\]?)|(?:[a-zA-Z0-9-._~!$&'()*+,;=]|%[0-9a-fA-F]{2})*))\5(?::(?=(\d*))\6)?)(\/(?=(("
+            r"?:[a-zA-Z0-9-._~!$&'()*+,;=:@\/]|%[0-9a-fA-F]{2})*))\8)?|(\/?(?!\/)(?=((?:[a-zA-Z0-9-._~!$&'()*+,"
+            r";=:@\/]|%[0-9a-fA-F]{2})*))\10)?)(?:\?(?=((?:[a-zA-Z0-9-._~!$&'()*+,;=:@\/?]|%[0-9a-fA-F]{2})*))\11)?("
+            r"?:#(?=((?:[a-zA-Z0-9-._~!$&'()*+,;=:@\/?]|%[0-9a-fA-F]{2})*))\12)?$")
```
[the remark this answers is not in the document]
mdazam1942: as discussed this regex can be used instead of url parse

stix_shifter_modules/gcp_chronicle/configuration/lang_en.json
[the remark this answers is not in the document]
lakshmi51974368: Updated the description as per the above comment

stix_shifter_modules/gcp_chronicle/stix_transmission/api_client.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class APIClient:
+    QUERY_ENDPOINT = 'v2/detect/rules'
+    QUERY_SCOPES = ['https://www.googleapis.com/auth/chronicle-backstory']
+    URI = "https://oauth2.googleapis.com/token"
+
+    def __init__(self, connection, configuration):
+        self.auth = configuration.get('auth')
+        self.auth['private_key'] = connection.get('selfSignedCert').replace('\\n', '\n')
```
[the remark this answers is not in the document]
lakshmi51974368: The original private_key comes with \n. It can be either in single line with \n or it can come with newline also. In order to handle private_key with or without \n, this replace statement is required.

stix_shifter_modules/gcp_chronicle/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,579 @@
...
+    @staticmethod
+    def _parse_time_range(qualifier, time_range):
+        """
+        Converts qualifier to timestamp format
+        :param qualifier: str
+        :param time_range: int
+        return: list of formatted timestamps
+        """
+        try:
+            compile_timestamp_regex = re.compile(START_STOP_PATTERN)
```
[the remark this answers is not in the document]
lakshmi51974368: Changes has been done to handle the qualifier without milliseconds as well.

stix_shifter_modules/gcp_chronicle/stix_transmission/ping_connector.py
lakshmi51974368: Updated code to handle the invalid private key with proper error code.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,374 @@
...
+        :param offset: str, offset value
+        :param length: str, length value
+        :return: dict
+        """
+        return_obj = {}
+        response_dict = {}
+        result = []
+        next_page_token = '0'
+        result_count = 0
+        page_size = self.api_client.result_limit
```
mdazam1942, line 32: why does the `page_size` set  to `result_limit`? shouldn't this be set to `length`?

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,374 @@
...
+        Fetching the results using search id, offset and length
+        :param search_id: str, search id generated in transmit query
+        :param offset: str, offset value
+        :param length: str, length value
+        :return: dict
+        """
+        return_obj = {}
+        response_dict = {}
+        result = []
+        next_page_token = '0'
```
mdazam1942, line 30: I see `next_page_token` is hardcoded and it never gets updated. why is it always `0`?

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,374 @@
...
+                response_dict['message'] = exp
+            self.logger.error('error when getting search results: %s', exp)
+            ErrorResponder.fill_error(return_obj, response_dict, ['message'], connector=self.connector)
+
+        finally:
+            if 'code' in response_dict:
+                if response_dict['code'] not in (1010, 1015):
+                    self.api_client.delete_search(search_id)
+            else:
+                self.api_client.delete_search(search_id)
```
mdazam1942, line 110: why are you deleting the search after every results call? I thought we decided to delete when the actual delete operation is called. 

Also, Results call returns error if the search is deleted. this happens when we run the results call again with the same search ID to retrieve the remaining results.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
lakshmi51974368: As discussed, updated the code to set the page_size with total of length and offset.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
lakshmi51974368: As discussed, updated the code to support with or without metadata in create_results_connection. IBM informed Metadata is not supported in isc-udi-connector 1.10.3.0. Once future versions supports metadata, the code should handle automatically.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
lakshmi51974368: Delete API call is added in order to avoid the overload of the chronicle database. Even if the users doesn't have permission to make the deletion API call, the code will return the results without exception.
The Deletion API call will be done only when all the required records are fetched or when next_page_token is not available(Detailed information has been sent in email).

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,374 @@
...
+                                                          connector=self.connector)
+                                return_obj = self.invalid_response(return_obj, response_dict,
+                                                                   next_response[0].status, response_text)
+                                break
+                        else:
+                            break
+                return_obj['success'] = True
+                if result:
+                    final_result = ResultsConnector.get_results_data(result)
+                    return_obj['data'] = final_result[int(offset):(int(offset) + int(length))] if final_result else []
```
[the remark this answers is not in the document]
lakshmi51974368: Updated code to have a new variable called total_records which is set based on offset and length.

stix_shifter_modules/gcp_chronicle/stix_transmission/api_client.py
```diff
@@ -149,7 +149,12 @@ def get_search_results(self, search_id, next_page_token, page_size):
 
         list_detection_endpoint = self.host + "/" + list_detection
 
-        return self.http_client.request(list_detection_endpoint, 'GET')
+        response = self.http_client.request(list_detection_endpoint, 'GET')
+
+        while response[0].status == 429:  # add sleep of 1 sec when resource exhaustion happens
+            time.sleep(1)
```
mdazam1942, line 155: are you sure sleeping 1 second resolve the resource exhaustion? I don't think you need to sleep if api returns 429. if resource is exhausted then just return whatever response you have got from the api.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -103,19 +139,25 @@ def create_results_connection(self, search_id, offset, length):
...
 
         finally:
-            if 'code' in response_dict:
-                if response_dict['code'] not in (1010, 1015):
-                    self.api_client.delete_search(search_id)
-            else:
-                self.api_client.delete_search(search_id)
+            try:
+                if not return_obj.get('data') or result_count >= self.api_client.result_limit or \
+                        response_dict.get('message') or not response_text.get('nextPageToken'):
```
mdazam1942, line 144: can you also put a debug log here about deleting the search?

lakshmi51974368: Updated the code with logger.debug when deletion is done in results_connector.py

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,418 @@
...
+        return events
+
+    @staticmethod
+    def validate_guardduty_event(events):
+        """
+        validate guardduty event
+        :param events: dict
+        :return: dict
+        """
+        # change the target location name to principal location name if resource role is target in AWS Guardduty
```
mdazam1942, line 364: this function is validating a very specific aws event. just wondering why is that? what if the product name is different? are you considering any other product type?

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -0,0 +1,418 @@
...
+        """
+        formats the results data
+        :return dict
+        """
+        for noun in ('principal', 'src', 'target'):
+            if noun in events['event'].keys():
+                events = ResultsConnector.format_registry_result(events, noun)
+                events = ResultsConnector.format_software_result(events, noun)
+                events = ResultsConnector.format_user_account(events, noun)
+        events = ResultsConnector.validate_guardduty_event(events)
```
mdazam1942, line 213: why do you need to validate guardduty events? what happens if you don't?

stix_shifter_modules/gcp_chronicle/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,1702 @@
...
+          "key": "file.modified",
+          "object": "about_file"
+        },
+        "fileType": {
+          "key": "file.extensions.x-gcp-chronicle-file.file_type",
+          "object": "about_file"
+        }
+      }
+    },
+    "additional": {
```
mdazam1942, line 1679: are these fields only specific to aws events? is chronicle always receives logs from aws? these are very odd extension properties.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
lakshmi51974368: This function validates AWS Guardduty event.  During validation of multiple log sources for the asset location, we found that the value of the source location name  is mapped to the target location name in UDM schema for AWS Guardduty events if the resource_role field in AWS is target. Hence we have added this specific code for Guardduty.
In other product types, we did not find any changes.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
lakshmi51974368: If the guardduty event is not validated, it would provide incorrect source location results in the final stix output.

stix_shifter_modules/gcp_chronicle/stix_translation/json/to_stix_map.json
lakshmi51974368: These fields are specific to AWS.  The “additional” key which belongs to UDM schema, doesn’t have any standard fields . It may vary between log sources.  Since observed this additional information in AWS events which had extra information like detector id and image id, we have added this information in to_stix mapping alone. If additional information needs to be added in future other log sources also, it  can add here.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
mdazam1942: just wondering, why is this specific to guardduty events? have you checked any other types of AWS events/logs? I assume all aws events set the location in a similar way.

lakshmi51974368: We had few logs from following AWS services other than Guardduty : VPC, Elastic load balancer, Cloud trial logs. During the coding part, we had validated these services logs.
But in the above mentioned services logs, after parsing into UDM schema, we did not find location name details to handle the region information. Hence we added this condition for Guardduty alone in results connector. 

Since this condition may create discrepancy and ambiguity between AWS Guardduty and other AWS services which may be impacted by this location attribute, we are removing this condition.
Hence the Stix output from to_stix mapping would come without any modifications for all AWS logs following its UDM schema conversion as such.

mdazam1942: location is basically a part of default UDM schema. I see,  the default parser parses location information of any product/datasource into UDM schema. In that case, we have to handle other product logs not just aws. For example CrowdStrike has region.

stix_shifter_modules/gcp_chronicle/stix_translation/json/to_stix_map.json
mdazam1942: This is fine for now.  we may have think about handling "additional" properties of other datasets from a different product.

mdazam1942 (approved):

PR #1078, opened 2022-08-12
Added Utility for normalization of connectors

SalmanMesia:
Added utility to create our own array of attributes from connectors

The attributes array can be used as a custom property in any SDO, that includes the name, value, and type of that attribute

### DIFF
--- a/stix_shifter/requirements.txt
+++ b/stix_shifter/requirements.txt
@@ -12,5 +12,3 @@ xmltodict==0.13.0
 jsonmerge==1.8.0
 colorlog==6.6.0
 stix2-matcher==3.0.0
-ipaddress==1.0.23
-validators==0.20.0
--- a/stix_shifter/requirements.txt
+++ b/stix_shifter/requirements.txt
@@ -12,3 +12,5 @@ xmltodict==0.13.0
 jsonmerge==1.8.0
 colorlog==6.6.0
 stix2-matcher==3.0.0
+ipaddress==1.0.23
+validators==0.20.0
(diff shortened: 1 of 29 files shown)


stix_shifter/requirements.txt
```diff
@@ -12,3 +12,5 @@ xmltodict==0.13.0
 jsonmerge==1.8.0
 colorlog==6.6.0
 stix2-matcher==3.0.0
+ipaddress==1.0.23
```
mdazam1942, line 15: ipaddress is part of standard python https://docs.python.org/3.9/library/ipaddress.html

stix_shifter_utils/normalization/normalization_helper.py
```diff
@@ -0,0 +1,64 @@
...
+  :return: ip if it is a string which is an ipaddress
+  :return: URI if it is a URL
+  :return: lat_lng if it is a latitude or longitude coordinates
+  '''
+  attribute_type = None
+  if isinstance(attribute, (int, float, complex)):
+    attribute_type = 'number'
+  if isinstance(attribute, (str)):
+    attribute_type = 'string'
+    if validators.url(attribute):
```
mdazam1942, line 56: validator library uses a whole bunch of regex underneath to validate. why don't you use `urllib.parse.urlparse`? you won't have to add any extra library to stix-shifter.

stix_shifter/requirements.txt
SalmanMesia: Removed

stix_shifter_utils/normalization/normalization_helper.py
SalmanMesia: Updated to remove validators and use urllib.parse instead

mdazam1942 (approved):

PR #1084, opened 2022-08-18
splunk: fix STIX timestamp processing

### LINKED ISSUE #855: Wrong start/stop qualifier format doesn't show proper error message in splunk
**Describe the bug**
If wrong timestamp format is specified in the start/stop qualifier for splunk translate command then it doesn't show proper error message.

**To Reproduce**

This splunk translate command:
`stix-shifter translate splunk query {} "[ipv4-addr:value = '10.1.1.1'] START t'2022-03-03T00:00:00Z' STOP t'2022-03-03T01:00:00Z'"`
yields the following error:
```
2022-03-17 08:44:27,593 stix_shifter.stix_translation.stix_translation ERROR Caught exception: Qualifier type not implemented <class 'NotImplementedError'>
2022-03-17 08:44:27,598 stix_shifter_utils.stix_translation.stix_translation_error_mapper ERROR received exception => NotImplementedError: Qualifier type not implemented
{
"success": false,
"code": "not_implemented",
"error": "wrong parameter : Qualifier type not implemented"
}
```
The error doesn't show whether the timestamp format is wrong.
**Expected behavior**

The error should indicate that the timestamp format is wrong.

**Additional context**
In general the\is timestamp should work but this is very specific to splunk. We need to verify why it is required to add `.SSS` in the. qualifier timestamp for splunk.

pcoccoli:
Fixes #855

### DIFF
--- a/stix_shifter_modules/splunk/stix_translation/query_constructor.py
+++ b/stix_shifter_modules/splunk/stix_translation/query_constructor.py
@@ -79,7 +79,6 @@ class SplunkSearchTranslator:
                     latest_obj = stix_strptime(latest)
                     latest_dt = latest_obj.strftime(splunk_date_format)
 
-                print('PC:', st_pattern, et_pattern, st_arr, et_arr, earliest, latest)
                 # prepare splunk SPL query
                 if earliest and latest:
                     return '{query_string} earliest="{earliest}" latest="{latest}"'.format(query_string=translated_query_str,
--- a/stix_shifter_modules/splunk/stix_translation/query_constructor.py
+++ b/stix_shifter_modules/splunk/stix_translation/query_constructor.py
@@ -13,6 +13,15 @@ from . import encoders
 from . import object_scopers
 
 
+def stix_strptime(date_string):
+    stix_date_format = "%Y-%m-%dT%H:%M:%S.%fz"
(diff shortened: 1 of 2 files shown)


stix_shifter_modules/splunk/stix_translation/query_constructor.py
```diff
@@ -43,31 +52,34 @@ def translate(self, expression, qualifier=None):
...
 
                 if et_arr:
                     # replace STOP and single quotes with empty char in date string
-                    latest = re.sub(r"(STOPt|')", '', et_arr[0] if et_arr else "")
-                    latest_obj = datetime.strptime(latest, stix_date_format)
+                    latest = re.sub(r"(STOPt|')", '', et_arr[0][0] if et_arr else "")
+                    latest_obj = stix_strptime(latest)
                     latest_dt = latest_obj.strftime(splunk_date_format)
 
+                print('PC:', st_pattern, et_pattern, st_arr, et_arr, earliest, latest)
```
mdazam1942, line 82: remove print. may be debug log if needed.

pcoccoli: Oops!

mdazam1942 (approved):

PR #1124, opened 2022-09-20
Get rid of StixObjectIdEncoder

baulus:


stix_shifter_utils/utils/helpers.py
```diff
@@ -1,16 +1,11 @@
...
-            return obj.object_id
-        # Let the base class default method raise the TypeError
-        return json.JSONEncoder.default(self, obj)
+import json_fix
 
 
 class StixObjectId(object):
     object_id:str = None
     def __init__(self, object_id):
+        json_fix
```
mdazam1942, line 8: any initialization value `json_fix` == ?

baulus: yes, it is a hack to make EDI and other developer (not to delete it when EDI suggests it is not used) to think it is used somehow, not just `import json_fix`
in reality just import json_fix is enough

baulus: having a second thought on this, I think you are right, any developer that sees this hack will think it is not initialized variable
removed, added comment next to import

delliott90 (approved):

baulus:
Fix link: https://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable#68926979
Lib: https://pypi.org/project/json-fix/#description

mdazam1942 (approved):

PR #1170, opened 2022-10-31
Upgrade pytests version for dev environment

mdazam1942:
upgrade fixes this alert: https://github.com/opencybersecurityalliance/stix-shifter/security/dependabot/1

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


delliott90 (approved):

delliott90:
Removal of Python 3.6 support in the documentation will be handled in https://github.com/opencybersecurityalliance/stix-shifter/pull/1159

mdazam1942:
> #1159

the latest versions of the updated package doesn't work in python 3.6 actually. that's why I had to remove it.

PR #1174, opened 2022-11-03
Updated to support self signed certificate authentication

lakshmi51974368:
Updated to support self signed certificate authentication.


stix_shifter_modules/rhacs/configuration/config.json
```diff
@@ -17,6 +16,14 @@
         "help": {
             "type": "link",
             "default": "data-sources.html"
+        },
+        "sni": {
+            "type": "text",
+            "optional": false
```
delliott90, line 22: I think this need to be optional true.

stix_shifter_modules/rhacs/configuration/config.json
```diff
@@ -17,6 +16,14 @@
...
+        },
+        "selfSignedCert": {
+            "type": "password",
+            "optional": false
```
delliott90, line 26: I think this need to be optional true.

stix_shifter_modules/rhacs/stix_transmission/api_client.py
```diff
@@ -13,10 +13,13 @@ def __init__(self, connection, configuration):
...
             headers['Authorization'] = "Bearer " + self.auth.get('token')
         url_modifier_function = None
+        #Added self-signed certificate parameter for verification.
         self.client = RestApiClient(connection.get('host'),
                                     connection.get('port', None),
                                     headers,
-                                    url_modifier_function=url_modifier_function
+                                    url_modifier_function=url_modifier_function,
+                                    sni=connection.get('sni'),
+                                    cert_verify=connection.get('selfSignedCert', True)
```
delliott90, line 22: This connector use to work with a CA certificate. I think we want this to work with both a CA or a self-signed certificate.

stix_shifter_modules/rhacs/configuration/config.json
lakshmi51974368: Updated config as suggested.

stix_shifter_modules/rhacs/configuration/config.json
lakshmi51974368: Updated config as suggested.

stix_shifter_modules/rhacs/stix_transmission/api_client.py
lakshmi51974368: Updated the code to support CA and self-signed certificate. Since "sni" and "selfSignedCert" is optional, it is not required to be passed for trusted CA issued server certificate.

stix_shifter_modules/rhacs/configuration/config.json
```diff
@@ -1,8 +1,7 @@
 {
     "connection": {
         "type": {
-            "displayName": "Red Hat Advanced Cluster Security for Kubernetes (StackRox)",
-            "group": "rhacs"
+            "displayName": "Red Hat Advanced Cluster Security for Kubernetes (StackRox)"
```
delliott90, line 4: Can you keep `"group": "rhacs"` in the config? It's used for identifying connectors for the same group of products that could share connection information.

stix_shifter_modules/rhacs/configuration/lang_en.json
```diff
@@ -1,24 +1,32 @@
...
+      "label": "Need additional help?",
+      "description": "More details on the data source setting can be found in the specified link"
+    },
+    "sni": {
+      "label": "Server Name Indicator",
+      "description": "The Server Name Indicator (SNI) enables a separate hostname to be provided for SSL authentication"
+    },
+    "selfSignedCert": {
+      "label": "RHACS Connection Certificate",
+      "description": "Use self-signed SSL certificate and CA content(root and intermediate) of data source"
```
delliott90, line 21: Wouldn't this be SSL certificate or CA certificate? The wording makes it sounds like both are suppose to be included here.

lakshmi51974368: Updated the text with "or" for better understanding.

stix_shifter_modules/rhacs/configuration/lang_en.json
```diff
@@ -1,24 +1,32 @@
...
-            "description": "Set the port number that is associated with the Host name or IP"
-        },
-        "help": {
-            "label": "Need additional help?",
-            "description": "More details on the data source setting can be found in the specified link"
-        }
+  "connection": {
+    "host": {
+      "label": "Management IP address or Hostname",
+      "description": "Specify the IP address or  hostname of the data source so that IBM Cloud Pak for Security can communicate with it"
```
delliott90, line 5: Please remove "so that IBM Cloud Pak for Security can communicate with it". These should be generic instructions since you would need this regardless of using in CP4S.

lakshmi51974368: Updated hostname description as suggested.

delliott90 (approved):

PR #1191, opened 2022-11-15
elastic_ecs: wrap MATCHES regex in forward slashes

### LINKED ISSUE #1190: elastic_ecs MATCHES operator is translated incorrectly
**Describe the bug**
The MATCHES operator doesn't work with elastic_ecs at all.  I believe the problem is that the module uses Elasticsearch "query string" syntax (this doesn't appear to be documented anywhere however) and regular expressions should be wrapped in `/`s: https://www.elastic.co/guide/en/elasticsearch/reference/8.5/query-dsl-query-string-query.html#_regular_expressions

There is code in the module that partly implements this but it's commented out: https://github.com/opencybersecurityalliance/stix-shifter/blob/119dd79f46ada08766f23505faaf4f263f9aa336/stix_shifter_modules/elastic_ecs/stix_translation/query_constructor.py#L112
```
            # if expression.comparator == ComparisonComparators.Matches:  # needs forward slashes
            #    value = self._format_match(expression.value)
            # should be (x, y, z, ...)
```

**To Reproduce**
Try any pattern with MATCHES:
```$ stix-shifter translate elastic_ecs query '{}' "[process:name MATCHES '.*\\.exe']"
line 1:22 mismatched input ''' expecting StringLiteral
 2022-11-14 16:38:39,472 stix_shifter_modules.elastic_ecs.stix_translation.query_translator INFO     Converting STIX2 Pattern to data source query
{
    "queries": [
        "(process.name : .*\\\\.exe OR process.parent.name : .*\\\\.exe) AND (@timestamp:[\"2022-11-14T21:33:39.473Z\" TO \"2022-11-14T21:38:39.473Z\"])"
    ]
}

```

**Expected behavior**
The translated query should wrap the RE with forward slashes.

**Additional context**
```
$ pip list | grep stix-shifter
stix-shifter                      4.5.1
stix-shifter-modules-elastic-ecs  4.5.1
stix-shifter-modules-qradar       4.5.1
stix-shifter-modules-splunk       4.5.1
stix-shifter-utils                4.2.4
$ pip list | grep antlr
antlr4-python3-runtime            4.8
```

pcoccoli:
Fixes #1190 
See https://www.elastic.co/guide/en/elasticsearch/reference/8.5/query-dsl-query-string-query.html#_regular_expressions


stix_shifter_modules/splunk/stix_translation/query_constructor.py
```diff
@@ -79,6 +79,7 @@ def translate(self, expression, qualifier=None):
                     latest_obj = stix_strptime(latest)
                     latest_dt = latest_obj.strftime(splunk_date_format)
 
+                print('PC:', st_pattern, et_pattern, st_arr, et_arr, earliest, latest)
```
mdazam1942, line 82: please remove the print statement.

mdazam1942 (approved):
The changes look fine. We don't have any elastic datasource to test. We can merge this if you confirm this works fine. @pcoccoli

pcoccoli:
I'm not sure it's possible to make elasticsearch regular expressions work at all, honestly, since hardly anything is supported.

pcoccoli:
Regular expression won't work as expected for fields like `process:command_line` (arguably the field for which they're needed the most) since they seem to be applied to the individual "terms" of the field.  It works for simple REs, e.g. `[process:name MATCHES 'cmd\\..*']` will map to `process.name: /.md.exe/`, and will match "cmd.exe".

Issue #1198, opened 2022-11-19
STIX 


delliott90:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

thangaraj-ramesh:
Added support for query with and without milliseconds in start and stop qualifier. #1199

delliott90:
Closed via https://github.com/opencybersecurityalliance/stix-shifter/pull/1199

PR #1199, opened 2022-11-21
Updated to support query without milliseconds in darktrace connector

thangaraj-ramesh:
Updated to support query without milliseconds in darktrace connector. #1198


thangaraj-ramesh:
This fix addresses the issue #1198

delliott90 (approved):

PR #1208, opened 2022-11-23
elastic_ecs: add beats dialect

pcoccoli:
Add a new "elk" dialect to the elastic_ecs connector.  This dialect is need for ELK deployments where the elasticsearch field mappings differ from what's documented in the ECS docs.
For example, `process.command_line` in ECS is documented here: https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-command-line
This states it's a wildcard field and has a `.text` subfield. 
However, actual deployments of ELK have mappings like this:
```
            "command_line" : {
              "type" : "text",
              "fields" : {
                "keyword" : {
                  "type" : "keyword",
                  "ignore_above" : 256
                }
              }
            },
```
Here, `process.command_line` is type "text" with a subfield of `.keyword` with type "keyword".  We actually want that subfield, since "text" fields are analyzed and don't work as STIX patterns require.

This "elk" mapping simply adds `.keyword` to the appropriate fields.


pcoccoli:
I should probably rename this dialect to "beats" as it winlogbeat, etc. that actually use it.

stix_shifter_modules/elastic_ecs/entry_point.py
```diff
@@ -11,5 +14,15 @@ def __init__(self, connection={}, configuration={}, options={}):
...
-        self.setup_translation_simple(dialect_default=dialect, results_translator=ResultTranslator(options, dialect))
\ No newline at end of file
+        query_translator = QueryTranslator(options, dialect, filepath)
+        results_translator = ResultTranslator(options, dialect)
+        self.add_dialect(dialect, query_translator=query_translator, results_translator=results_translator, default_include=True, default=True)
+
+        dialect = 'beats'
+        query_translator = QueryTranslator(options, dialect, filepath)
+        results_translator = ResultTranslator(options, dialect)
+        self.add_dialect(dialect, query_translator=query_translator, results_translator=results_translator, default_include=False, default=False)
```
mdazam1942, line 28: I think default `from_stix_map.json` should be renamed to `ecs_from_stix_map.json`. that way we will have two dialects ecs and beats.

Another thing, should this always construct two queries? if not unless user sets the beats dialects through datasource settings, it will always construct default query which is ecs in this case.

mdazam1942 (commented):
I think default from_stix_map.json should be renamed to ecs_from_stix_map.json. that way we will have two dialects `ecs` and `beats`.

Another thing, should this always construct two queries? if not unless user sets the beats dialects through datasource settings, it will always construct default query which is ecs in this case.

stix_shifter_modules/elastic_ecs/entry_point.py
mdazam1942, line 28: if two dialects are created then you should be able to use setup_translation_simple() -

`self.setup_translation_simple(dialect_default='ecs')`

pcoccoli: Are you sure?  I tried using that and it wasn't working.

Is this required to merge?  I don't really have time to work on this further.

mdazam1942: it should work. we already have few connectors that uses the similar structure. Here are couple of examples:

https://github.com/opencybersecurityalliance/stix-shifter/blob/61e8975c39836981c61ce3f6cbd1f8bf6e2eedba/stix_shifter_modules/aws_athena/entry_point.py#L29
https://github.com/opencybersecurityalliance/stix-shifter/blob/61e8975c39836981c61ce3f6cbd1f8bf6e2eedba/stix_shifter_modules/qradar/entry_point.py#L12

This is just my suggestion to make the module structure consistent. We can do this later if you dont have time atm.

mdazam1942:
@pcoccoli if you have nothing else to update. we can merge this pr. please confirm.

pcoccoli:
I have no more updates; thanks.

PR #1239, opened 2022-12-13
mapping fixes for splunk

delliott90:


pcoccoli (approved):
I think this looks like an improvement; I'm just not sure about process_path.  The CIM docs show that as the absolute path of the binary/image (including the filename), whereas we just want the directory path.

mdazam1942 (approved):

PR #1266, opened 2023-01-04
Removed language common fields

baulus:


baulus:
This is not going to go to develop for now.

PR #1291, opened 2023-01-24
fix(CP4SCORE-18414): fix wrong host mapping

JingqiuDu:


stix_shifter_modules/reaqta/stix_translation/json/from_stix_map.json
```diff
@@ -40,7 +40,7 @@
   },
   "process": {
     "fields": {
-      "binary_ref.name": ["filename"],
+      "image_ref.name": ["filename"],
```
delliott90, line 43: This isn't the file you want to change this. This is for STIX 2.0 in which case `binary_ref` is correct. This change should happen in the STIX 2.1 mapping files., both the from and to-stix mappings.

JingqiuDu: Fixed

delliott90 (approved):

Issue #1285, opened 2023-01-19
stix-shifter return records have limited size


tingdai:
**Describe the bug**
- When executing a query that requires more than 10k returns, stix-shifter failed with RuntimeError.

**To Reproduce**
- Use stix-shift to execute a query and specify the result limit more than 10000, e.g., `--results 11000`. An example command shows as follows:
```
stix-shifter execute elastic_ecs elastic_ecs '{"type":"identity","id":"xxx","name":"elastic","identity_class":"events"}' '{"host":"xxx", "port":xxx, "selfSignedCert": false}' '{"auth":{"id": "xxx", "api_key":"yyy"}}' "[ipv4-addr:value = '127.0.0.1'] START t'2021-10-17T12:00:00Z' STOP t'2022-10-18T00:00:00Z'" --results 11000
```

**Expected behavior**
On ElasticSearch side, the page limit is set as 10k by default, when stix-shifter wants to retrieve more than 10k data using simple [GET API](https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_modules/elastic_ecs/stix_transmission/api_client.py#L55), RuntimeError happens. 

**Additional context**
As I[ssue 143](https://github.com/opencybersecurityalliance/kestrel-lang/issues/143) listed, we need to use advanced paging API to transmit results between stix-shifter and ElasticSearch.

tingdai:
For large queries, we can use [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html) API as suggested.
However, there are some challenges that need to addressed:
1. We cannot use indices (plural) for point in time ([PIT](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html)). This is because `search_after` does not support it. For example, if we use the following APIs: 
```
POST winlogbeat-*/_pit?keep_alive=2m

GET winlogbeat-*/_search?size=10&from=0
{
  "query": {
    "query_string": {
      "query": "xxxxx"
    }
  },
  "pit": {
    "id": "xxxxx",  ### the id is provided by the previous POST api
    "keep_alive" : "2m"
  },
  "sort": [
    {"@timestamp": "asc"}
  ]
}
```
we will  get validation failures:
```
Validation Failed: 1: [indices] cannot be used with point in time. Do not specify any index with point in time.;
```
[...]

PR #1299, opened 2023-01-27
support large query with elastic search_after pagination

tingdai:
Use pagination to support large queries when the size is >10 k (default `max_result_window` in ElasticSearch).
This is related to https://github.com/opencybersecurityalliance/stix-shifter/issues/1285


tingdai:
Reimplement ElasticSearch pagination by moving while loop from stix-shifter to Kestrel in 1e89e283be2134a5d417301e0f872d4c63eb0f63

mdazam1942 (changes requested):
@tingdai can you please fix the unittests?

tingdai:
> @tingdai can you please fix the unittests?

@mdazam1942 I fixed testcase issues with the following commits: fe5ed4c38ecfe5defabe112d5c3c843b331d3563 and 2573ef6d539ea7d76bf030d738dd8918a5bc489e

stix_shifter_modules/elastic_ecs/stix_transmission/connector.py
```diff
@@ -44,12 +52,44 @@ def ping_connection(self):
...
+                self.logger.error('can not parse response: ' + str(response_txt))
+            else:
+                raise e
+
+    def create_results_connection(self, query, offset, length, metadata=None):
         response_txt = None
         return_obj = dict()
 
         try:
-            response = self.api_client.run_search(query, offset, length)
```
mdazam1942, line 52: so if api_client.run_search() never going to be used. can we just remove run_search() function from api client?

stix_shifter_modules/elastic_ecs/stix_transmission/connector.py
```diff
@@ -44,12 +52,44 @@ def ping_connection(self):
...
+            # using search after API in ElasticSearch
+            # pass the last searched value in metadata argument, ignore offset argument
+            response = self.api_client.search_pagination(query, metadata, min(length, self.max_result_window))
```
mdazam1942, line 92: i would cast length into int() to make sure it doesn't fail at this call:` int(length)`

mdazam1942: passing length in CLI arguments fails at this point because it reads it as a string.

stix_shifter_modules/elastic_ecs/stix_transmission/connector.py
tingdai: Yes, I guess we can remove `run_search()` function.

delliott90:
There is one failing unit test because the `run_search()` function was removed.

stix_shifter_modules/elastic_ecs/stix_transmission/connector.py
```diff
@@ -59,11 +99,29 @@ def create_results_connection(self, query, offset, length):
                     self.logger.error("Total # of hits:" + str(response_json['hits']['total']))
                     return_obj['data'] = [record['_source'] for record in response_json["hits"]["hits"]]
                     self.logger.error("Total # of records: " + str(len(return_obj['data'])))
-
+                    if len(response_json["hits"]["hits"]) == 0:
+                        return_obj['lastsort'] = metadata
+                    elif 'sort' in response_json["hits"]["hits"][-1]:
+                        return_obj['lastsort'] = response_json["hits"]["hits"][-1]['sort']
```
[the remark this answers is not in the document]
tingdai: I don't think we can change it now because Kestrel currently uses `lastsort`.

tingdai:
> There is one failing unit test because the `run_search()` function was removed.

@delliott90 Fixed in the new commit d0571da25dfe8d461326f9aff6fa1d0d921460cb

stix_shifter_modules/elastic_ecs/stix_transmission/connector.py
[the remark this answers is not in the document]
tingdai: committed in 87833349d3ebe89bce70d4df1063a907a53b670c

delliott90 (approved):

mdazam1942 (approved):

PR #1323, opened 2023-02-13
Okta UDI connector

thangaraj-ramesh:
Changes in json_to_stix_translator.py file for grouping the references of list of dictionary attribute in final stix results


stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -242,6 +242,11 @@ def _handle_properties(self, to_stix_config_prop, data, objects, object_tag_ref_
                             # data variable is the final value, process in bulk
                             self._handle_value(data, parent_data, ds_sub_key, to_stix_config_prop, objects, object_tag_ref_map, object_key_ind)
                             break
+                    # group the references of list of dictionary field
+                    if isinstance(to_stix_config_prop, dict) and to_stix_config_prop.get('groupReference'):
+                        self._handle_value(data, to_stix_config_prop, ds_sub_key,
+                                           to_stix_config_prop['groupReference'],
+                                           objects, object_tag_ref_map, object_key_ind)
```
baulus, line 249: Excellent find! 
I tried to find a better place for this code, but for now it seems the most appropriate place. 
Can we do it more universal, without hardcoding? Something like this:
```
                    # group the references of list of dictionary field
                    if isinstance(to_stix_config_prop, dict):
                        group_refs = [key for key, value in to_stix_config_prop.items() if isinstance(value, dict) and value.get('group_ref')]
                        if group_refs:
                            for group_ref in group_refs:
                                self._handle_value(data, to_stix_config_prop, ds_sub_key, to_stix_config_prop[group_ref], objects, object_tag_ref_map, object_key_ind)
```

If so, you will need also to add `"group_ref": true` to groupReference property in the to_stix_map as well

baulus: Also, your unit tests pass both with and without this code you added, can you please add unit test case for this group reference?

thangaraj-ramesh: In our code we are planning to create list of dict as a separate objects and map its references as individual object reference and not at atttibute level.
as per our understanding of the above code, this tries to get the references at attribute level of the dictionary. so when we tried to verify the same, this creates duplicate objects with references. we have sent the detailed explanation in email.

baulus: right, it is possible, the meaning of my code is just to help to avoid hardcoding by property name (there can be others) by looking up a specific attribute, just like `wrap: true` we have. 
I will look at the email. thank you!

mdazam1942 (approved):

stix_shifter_modules/okta/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,318 @@
...
+      "key": "x-oca-event.outcome",
+      "object": "event"
+    },
+    "reason": {
+      "key": "x-oca-event.extensions.x-okta-event.outcome_reason",
+      "object": "event"
+    }
+  },
+  "legacyEventType": {
+    "key": "x-oca-event.extensions.x-okta-event.legacy_event_type",
```
mdazam1942, line 21: The extension mappings looks odd. we need to rethink how we should map the extensions. For this kind of mapping, we can simply add custom properties rather than putting in the extension.

mdazam1942 (commented):
PR looks good. But we need to discuss about the extension mappings: https://github.com/opencybersecurityalliance/stix-shifter/pull/1323/files#r1126912941
 Possibly clean this up before merging.

PR #1324, opened 2023-02-13
Update elastic_ecs_supported_stix.md

ionsor:
Small correction, possible due to mistype. AND operator in STIX it is AND in Data Source Operator

### DIFF
--- a/adapter-guide/connectors/elastic_ecs_supported_stix.md
+++ b/adapter-guide/connectors/elastic_ecs_supported_stix.md
@@ -3,7 +3,7 @@
 ### Supported STIX Operators
 | STIX Operator | Data Source Operator |
 |--|--|
-| AND | OR |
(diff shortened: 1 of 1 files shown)


ionsor:
My commit was unsuccessful, however I still recommend to correct this small type.

delliott90:
The supported mappings is actually generated by the following script based on the mapping files: https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter/scripts/supported_property_exporter.py
The issue is that the AND operator between observations `[ ]` is treated as an OR operator which is expected, but AND operators within the observation such work as AND operations. We will need to fix the script to account for this since the operator table is misleading right now.

delliott90:
Resolved with https://github.com/opencybersecurityalliance/stix-shifter/pull/1348.

PR #1342, opened 2023-02-23
Change to domain name mapping

KaneBrennan132:
Changed the domain-name mapping as DOMAINNAME(domainid) is for the Qradar tenancy domains.

User documentation changes are planned.


stix_shifter_modules/qradar/stix_translation/json/stix_2_1/to_stix_map.json
```diff
@@ -465,7 +465,9 @@
       "key": "domain-name.value",
       "object": "domain2",
       "transformer": "ToDomainName"
-    },
+    }
+  ],
+  "qradardomain": [
```
delliott90, line 470: You should remove the entry that's currently under `dnsdomainname` a few lines down. We don't want to write duplicated data in multiple properties. In fact, if dnsdomainname is truly a domain, it shouldn't be writing the value to the url object.

delliott90 (changes requested):
Any changes to the `to-stix` mapping files should be reflected in the `from-stix` files as well.

KaneBrennan132:
Changed dnsdomainname entry to only write to domain-name
Changed alias used by DNS request domain in event search to use the alias dnsdomainname, removed domainname entry as it is no longer required. 
Updated Unit Tests to reflect changes
Updated Docs to reflect changes

delliott90 (approved):

PR #1350, opened 2023-02-27
cybereason quick ping

yurii-klymenko:


stix_shifter_modules/cybereason/stix_transmission/connector.py
```diff
@@ -115,13 +115,15 @@ def ping_connection(self):
...
-            if response.response.history:                           # If the authentication is invalid, the history
-                if response.response.history[0].status_code == 302:  # will be returned with 302 status code.
-                    raise InvalidAuthenticationException
-            response_code = response.response.status_code
-            response_dict = json.loads(response.response.text)
-            if response_code == 200 and response_dict['status'] == 'SUCCESS':
+            response_code = response.code
+            if response_code == 302:
+                raise InvalidAuthenticationException
+            if response_code == 405:
```
mdazam1942, line 121: Setting 405 as success seems odd.

stix_shifter_modules/cybereason/stix_transmission/api_client.py
```diff
@@ -32,7 +32,7 @@ def ping_box(self):
         query = '{"queryPath": [{"requestedType": "Connection", "filters": [], "isResult": true}],' \
                 '"totalResultLimit": 1, "perGroupLimit": 1,"templateContext": "SPECIFIC"}'
         headers['Cookie'] = self.session_log_in()
-        return self.client.call_api(self.QUERY_ENDPOINT, 'POST', headers=headers, data=query)
+        return self.client.call_api(self.QUERY_ENDPOINT, 'HEAD', headers=headers, data=query)
```
mdazam1942, line 35: in this case, i think we dont even need to make any QUERY_ENDPOINT call.  if `self.session_log_in()` is success then we can set ping successful.

yurii-klymenko: there is no way to know if it is successful with underlying call, it always return something like:`JSESSIONID=281AD9E056714688EA6F3423AC21E`

mdazam1942 (approved):

PR #1352, opened 2023-03-01
QRadarEpochToTimestamp for exponential notation

KaneBrennan132:
When using "group by" during an AQL passthrough, it was found that Qradar can return epoch values using exponential notation (e.g 1.674125456158E12) and this would fail being transformed into a timestamp, causing the DE search to fail. 

Added a custom epoch to timestamp transformer that handles epoch values using exponential notation.


delliott90:
Can we get a unit test around this?

KaneBrennan132:
Added test for start and end time epoch values using exponential notation.

delliott90 (approved):

PR #1364, opened 2023-03-07
Metadata changes for GCP chronicle

thangaraj-ramesh:
1. Modified the metadata parameter type to dictionary and handled 429 resource exhaustion exception.
2. Added unit test cases and updated readme for the same.

### DIFF
--- a/stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
+++ b/stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
@@ -8,6 +8,10 @@ import copy
 import re
 
 
+class InvalidMetadataException(Exception):
+    pass
+
+
 class ResultsConnector(BaseResultsConnector):
     EMAIL_PATTERN = re.compile(r'(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)')
     GCP_MAX_PAGE_SIZE = 1000
@@ -23,7 +27,7 @@ class ResultsConnector(BaseResultsConnector):
         :param search_id: str, search id generated in transmit query
         :param offset: str, offset value
         :param length: str, length value
-        :param metadata: str
+        :param metadata: dict, result_count, next_page_token
         :return: dict
         """
         return_obj = {}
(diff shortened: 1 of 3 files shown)


stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
```diff
@@ -86,31 +93,56 @@ def create_results_connection(self, search_id, offset, length, metadata=None):
                                     local_result_count += len(response_text['detections'])
                                     result.append(response_text['detections'])
                             else:
-                                return_obj = self.invalid_response(return_obj, response_dict,
-                                                                   next_response[0].status, response_text)
-                                result = []
+                                # recursive calls to handle 429 resource exhaustion exception during pagination
+                                if next_response[0].status == 429:
```
mdazam1942, line 97: why do you need to specifically handle 429?  How would you know server accept the immediate next call? If the server is busy then you should return the error response unless this is a very special API call.

delliott90:
This should now need to be targeted to the `develop_4.x` branch and a new PR can be opened to bring this into develop.

stix_shifter_modules/gcp_chronicle/stix_transmission/results_connector.py
thangaraj-ramesh: We have observed GCP chronicle returns resource exhastion error (status code 429) when we try to run a query with limit more than 10K in CP4S and also when we run multiple queries in parallel. As a result of this error, we were not able to fetch results more than 10K and we received only partial results . On further anaysis , we have observed that there exists an api rate limit of 10 api hit per 60 seconds for results api(list detection api) which causes this resource exhaustion error. Inorder to fetch more results, we have specifically handled this status 429 exception in the connector code. 
In case of other status codes exceptions, corresponding exceptions are thrown in the CP4S.

In command line, we can search with a maximum length of 10K since the resource exhaustion (429 exception) is handled based on metadata value returned in CP4S loop which is not available in command line

thangaraj-ramesh: We have updated the base branch to develop_4.x. kindly let us know your comments.

thangaraj-ramesh: Updated code with proper error responses for 429 exception and recursion calls are removed.

mdazam1942 (approved):

mdazam1942 (approved):

PR #1392, opened 2023-03-16
get configs

yurii-klymenko:
python -m main configs
  or
stix-shifter configs

### DIFF
--- a/stix_shifter/scripts/stix_shifter.py
+++ b/stix_shifter/scripts/stix_shifter.py
@@ -75,10 +75,12 @@ def main():
     translate_parser.add_argument('-x', '--stix-validator', action='store_true',
                                   help='Run the STIX 2 validator against the translated results')
     # configs parser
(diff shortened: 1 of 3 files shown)


mdazam1942 (commented):
it should have the option to get the config of a specified module as well instead of just getting all the module configs. like this- 

`python main.py configs qradar`

yurii-klymenko:
it is implemented the same way as 
`python -m main modules`
so cli returns for all of them,
but in code you can do:
`config = await stix_tranlation.translate_async('qradar', CONFIGS, {}, {}, {})`
to fetch individual one

yurii-klymenko:
added:
`
python -m main modules 
python -m main modules qradar
python -m main configs
python -m main configs qradar
`

mdazam1942 (approved):
@yurii-klymenko thanks for adding those extra arguments. looks great

PR #1397, opened 2023-03-21
Msatp with alerts

aviv1ron1:
big upgrade to the msatp module. look at the readme file for all details.
- improve from stix mappings
- add two event tables to the mappings which were not being searched
- add alerts table
- improve to stix mappings


aviv1ron1:
obsolete, merged thru PR 1404 instead

PR #1419, opened 2023-04-04
Added async to Azure sentinal

baulus:


stix_shifter/requirements.txt
```diff
@@ -5,6 +5,7 @@ aiomysql==0.1.1
 antlr4-python3-runtime==4.8
 asyncio==3.4.3
 asynctest==0.13.0
+azure-identity==1.12.0
```
delliott90, line 8: This should go in it's own requirements file under the azure_sentinel module since it's connector-specific. It's likely something we should do with some of these other libraries as well.

baulus: At least 3 connectors are going to use it (msatp, azure sentinel and azure log), there might be more connectors that will be based on Microsoft API.
Same approach we have for aioboto3

delliott90: Yeah, that's fine then.

mdazam1942 (approved):

PR #1436, opened 2023-04-23
Update Reversinglabs connector

SalmanMesia:
Fixed defects
Added custom attributes to the Indicator SDO (threat score and threat attributes)


stix_shifter_modules/reversinglabs/stix_translation/query_constructor.py
```diff
@@ -19,9 +19,27 @@
 
 class QueryStringPatternTranslator:
     # Change comparator values to match with supported data source operators
+    comparator_lookup = {
```
delliott90, line 22: This shouldn't be here. All operators should be defined in the `operators.json` file.

delliott90 (approved):

PR #1443, opened 2023-04-27
DShield connector

SalmanMesia:
Introduces DShield Connector which queries IPv4, and IPv6 Indicators of Compromise to generate a STIX 2.1 threat report from the Threat Feed with Identity, Extension-Definition and Indicator SDOs

Depending on the response from DShield APIs, we might have additional SDOs in our STIX Bundle Response


stix_shifter_modules/dshield/.coveragerc
```diff
@@ -0,0 +1,2 @@
+[run]
+omit = tests/*
```
mdazam1942, line 2: @SalmanMesia i would move `.coveragerc` to root level. you probably have this in few modules. i would recommend to remove them and put `.coveragerc` in `stix-shifter/`

mdazam1942 (approved):

PR #1464, opened 2023-05-03
Elastic-ecs: Patch observer mapping to `x-oca-asset` object

Harmedox:
This PR fixes the mapping of observer data to `x-oca-asset`, testing the mapping accuracy and references.

### DIFF
--- a/stix_shifter_modules/elastic_ecs/stix_translation/json/stix_2_1/to_stix_map.json
+++ b/stix_shifter_modules/elastic_ecs/stix_translation/json/stix_2_1/to_stix_map.json
@@ -1913,11 +1913,15 @@
     "ip": [
       {
         "key": "ipv4-addr.value",
-        "object": "observer_ip"
+        "object": "observer_ip",
+        "unwrap": true,
+        "transformer": "FilterIPv4List"
       },
       {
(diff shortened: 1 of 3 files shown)


stix_shifter_modules/elastic_ecs/stix_translation/json/stix_2_1/to_stix_map.json
```diff
@@ -1910,14 +1910,42 @@
...
-      "object": "observer"
-    },
+    "ip": [
+      {
+        "key": "ipv4-addr.value",
+        "object": "observer_ip"
+      },
+      {
+        "key": "ipv6-addr.value",
+        "object": "observer_ipv6"
```
mdazam1942, line 1920: need to make sure if the value is ipv4 or ipv6 before setting the value. one IP can't be both ipv4 and v6. you may have to use a transformer to verify the value. i know few connectors has this confusions. but we are planning to fix them eventually.

Harmedox: fixed. I added the transformer for ipv4 and v6.

mdazam1942 (approved):

PR #1469, opened 2023-05-05
to_stix_map validator

pcoccoli:
A work in progress really, but should still be useful.

PR #1471, opened 2023-05-07
fix mapping references in elastic-ecs connector

Harmedox:
This PR fixes mapping errors identified by the [to_stix_map validator](https://github.com/opencybersecurityalliance/stix-shifter/pull/1469).


stix_shifter_modules/elastic_ecs/stix_translation/json/to_stix_map.json
```diff
@@ -221,8 +222,7 @@
       }
     },
     "address": {
-      "key": "x-ecs-client.address",
-      "object": "client"
+      "key": "x-oca-asset.address"
```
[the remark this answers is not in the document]
Harmedox: that was intentional. `address` is not currently mapped to any STIX attribute, hence no object. For clarification, the value in `address` is either represented as a domain or IP, both of which are already mapped to `x-oca-asset`.

stix_shifter_modules/elastic_ecs/stix_translation/json/to_stix_map.json
```diff
@@ -1015,8 +1089,8 @@
         }
       ],
       "ppid": {
-        "key": "process.parent_ref.ppid",
```
subbyte, line 1018: Not sure if we have `process:ppid` and `process:pgid` in STIX. May need a double check.

Harmedox: Both attributes are `pid` refs. `ppid` refers to the parent process `pid`, while `pgid` refers to the process group leader `pid`. NB: `pgid` to be deprecated and referred to as `process.group_leader.pid`. 

Fixed the refs in `to_stix_map`.  See new [commit](https://github.com/opencybersecurityalliance/stix-shifter/pull/1471/commits/a301c5d30f35b382aaeb7f5e1a1257918fdadac1).

A new `group_leader_ref` reference attribute should be added to [Process STIX extensions](https://github.com/opencybersecurityalliance/stix-extensions/blob/main/2.0/custom-process.md).

subbyte:
Quick check: in `to_stix`, does the `source` entry has `address` similar to `destination.address`? I could not find the field in the `develop` branch, just to double check if we have fixed it here.

mdazam1942 (approved):

PR #1486, opened 2023-05-16
Develop_4.x_Qradar_Filter_Zero_Values

KaneBrennan132:
Addition of checks for the JSON results for zero values amongst the IP and mac addresses as this data is irrelevant it should be filtered out in order to provide more correct results in smaller size. Sets them to None.

Addition of Unit test for theses values.

### DIFF
--- a/stix_shifter_modules/qradar/stix_translation/results_translator.py
+++ b/stix_shifter_modules/qradar/stix_translation/results_translator.py
@@ -21,6 +21,34 @@ class ResultsTranslator(JSONToStix):
       
       if result.get('flowdestinationpayload'):
         result['mime_type_flowdestinationpayload'] = 'application/octet-stream'
+      
+      if result.get('sourceip'):
+        if result['sourceip'] == '0.0.0.0':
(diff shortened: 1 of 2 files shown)


stix_shifter_modules/qradar/stix_translation/results_translator.py
```diff
@@ -21,6 +21,34 @@ def translate_results(self, data_source, data):
       
       if result.get('flowdestinationpayload'):
         result['mime_type_flowdestinationpayload'] = 'application/octet-stream'
+      
+      if result.get('sourceip'):
+        if result['sourceip'] == '0.0.0.0':
+          result['sourceip'] = None
```
mdazam1942, line 27: can you use a transformer for this case instead of hard coding the values?

KaneBrennan132: Hi mdazam1942, these changes where already approved and merged into Develop by Danny. This is is for adding them into the Develop_4.x.

delliott90 (approved):

PR #1501, opened 2023-06-01
Patch elastic mappings

Harmedox:
Make the `id`, `type`, and `name` fields of `x-oca-asset` consistent with the STIX attributes specified in https://github.com/opencybersecurityalliance/stix-extensions/blob/main/2.0/x-oca-asset.md.

In details:
1. replace the reserved keywords `id` and `type` in `from_stix` with `device_id` and `host_type` respectively.
2. merge the `name` attribute into `hostname`.


stix_shifter_modules/elastic_ecs/stix_translation/json/beats_from_stix_map.json
```diff
@@ -184,7 +184,7 @@
       "severity": ["event.severity"],
       "start": ["event.start"],
       "timezone": ["event.timezone"],
-      "type": ["event.type.keyword"],
```
subbyte, line 187: Do we have `event_type` in `x-oca-event`, or `category`?

subbyte: Shall we also check whether we can use the `category` attribute?

Harmedox: By definition, it’s not the same thing. But considering `category` is an array of string, it can be used. How frequently is category used in practice?

subbyte: I think we don't have statistics about the popular data regarding `event.type`. We may raise this as a question tomorrow and get others' thoughts and understandings on `category`.

Harmedox: The [kind](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-kind.html), [type](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html) and [category](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-category.html) fields of the event category hierarchy have been consolidated in `x-oca-event:category`. See [0bae039](https://github.com/opencybersecurityalliance/stix-shifter/pull/1501/commits/0bae039195ea945d9ca27477c4601db150f19300).

mdazam1942 (approved):

PR #1502, opened 2023-06-05
remove cybox false flag for observed-data properties

delliott90:


stix_shifter_utils/stix_translation/src/json_to_stix/json_to_stix_translator.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, data_source, ds_to_stix_map, transformers, options, callback=
         self.callback = callback
 
         # parse through options
-        self.cybox_default = options.get('cybox_default', True)
+        # self.cybox_default = options.get('cybox_default', True)
```
mdazam1942, line 69: this line is not needed anymore. do you want to keep it for future reference since you commented out?

mdazam1942 (approved):

PR #1519, opened 2023-06-26
Move results processing to transmission results

mdazam1942:


delliott90:
If all the results_translators are being removed, shouldn't it's reference also be removed from the base entry point? https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_utils/utils/base_entry_point.py#L126

delliott90:
You will likely want to remove the `results_translator` from the various template projects.

PR #1529, opened 2023-07-18
To-STIX mapping keyword documentation

mdazam1942:


delliott90:
You will likely want to link the keywords page from https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/adapter-guide/develop-translation-module.md
Also, it may be helpful to add a column in the table showing the STIX output.

delliott90 (approved):

PR #1530, opened 2023-07-19
Vectra UDI connector

thangaraj-ramesh:
Adding Vectra UDI connector


stix_shifter_modules/vectra/stix_transmission/connector.py
```diff
@@ -0,0 +1,215 @@
...
+        return_obj = {}
+        data = []
+        try:
+            offset = int(offset)
+            length = int(length)
+            pagination_offset = offset
+            total_records = length
+            page = 1
+
+            # changing offset to page number
```
mdazam1942, line 54: can you explain this calculation? what is `pagination_offset`? why does it needs to be less than 5? what is the difference between page and page_size?

stix_shifter_modules/vectra/stix_transmission/connector.py
```diff
@@ -0,0 +1,215 @@
...
+                    next_url = next_response.get('next')
+                    if response_code == 200:
+                        data += next_response['results']
+                    else:
+                        response = next_response
+                        break
+
+                if response_code == 200:
+                    return_obj['success'] = True
+                    return_obj['data'] = self.get_results_data(data[pagination_offset:total_records])
```
mdazam1942, line 97: why do you need to slice of the results list again since you are already checking the results count inside while loop?

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: Previous page calculation corrected, please check the below example 

Vectra doesn't support offset values in queries, we have the option to mention page(page number) and page size.
We need to find out given offset is under which page and then we will fetch the particular page.
example: inputs offset=2010 and length=2000
		converting offset 2010 to page = 2, page_size = 2000 and pagination_offset = 10				
		Page and page_size(length) are API parameters for the query.
		pagination_offset is need for slicing.
		
Here, offset 2010 is under page 2. So, will fetch the page 2, which contains 2000 records.
From the records, we are slicing from 10(pagination_offset) to 2000.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: Please check the above comment; it explains slicing.

stix_shifter_modules/vectra/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,678 @@
...
+from datetime import datetime, timedelta
+from stix_shifter_utils.utils import logger
+from stix_shifter_utils.stix_translation.src.patterns.pattern_objects import ObservationExpression, \
+    ComparisonExpression, ComparisonComparators, Pattern, CombinedComparisonExpression, CombinedObservationExpression
+
+logger = logger.set_logger(__name__)
+
+START_STOP_PATTERN = r"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?Z)"
+STOP_TIME = datetime.utcnow()
+DEFAULT_LIMIT = 10000
```
mdazam1942, line 14: what is default limit? why is it `10000`?

stix_shifter_modules/vectra/stix_translation/query_constructor.py
```diff
@@ -0,0 +1,678 @@
...
+            raise RuntimeError(f'Unknown Recursion Case for expression={expression}, '
+                               f'type(expression)={type(expression)}')
+
+    def parse_expression(self, pattern: Pattern) -> list:
+        """
+         Formation of vectra query from ANTLR parsing expression.
+        :param pattern: expression object, ANTLR parsed expression object
+        """
+        query = self._parse_expression(pattern)
+        if self.options['result_limit'] > DEFAULT_LIMIT:
```
mdazam1942, line 415: is there a limitations on the result limit? not sure why you are replacing result_limit to  `DEFAULT_LIMIT`

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: fair calculation. but just wondering, does `next_url` contains page,page_size for next set of results? I wonder if you can make use of next_url to retrieve next batch of results in the subsequent calls.

stix_shifter_modules/vectra/stix_transmission/connector.py
```diff
@@ -0,0 +1,212 @@
...
+            response_code = response_wrapper.code
+            response = json.loads(response_wrapper.read().decode('utf-8'))
+
+            if response_code == 200:
+                data += response['results']
+                next_url = response.get('next')
+                while next_url:
+                    if len(data) >= total_records:
+                        break
+                    next_url = next_url.split('/api/v2.4/search/detections/?')
```
mdazam1942, line 80: why do you need to split() the next_url? you can just directly make api call using the next_url instead of this extra steps.

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
```diff
@@ -0,0 +1,1216 @@
+{
+  "src_ip": [
```
mdazam1942, line 2: is `src_ip` not associated with any network-traffic objcet? i see dst_ip is mapped to dst_ref for network-traffic.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: yes, next_url contains query, page, and page_size.

stix_shifter_modules/vectra/stix_translation/query_constructor.py
thangaraj-ramesh: removed the default limit; it's not required.

stix_shifter_modules/vectra/stix_translation/query_constructor.py
thangaraj-ramesh: removed the default limit; it's not required.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: next_url contains host, path, and query with pagination parameters.
self.api_client.get_search_results function expects only query, so we are splitting the next_url and passing the query with pagination parameters.
Host and path details were already configured in the api_client.py file, it requires only a query input.
Reuse the function, we are splitting the next_url.

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
thangaraj-ramesh: All network-traffic objects are constructed from the nested list in the source data while as src_ip is mapped outside the list. Framework doesn't support referencing src_ip into the nested list as reference is index based. For a particular record there is only a single host which is compromised and this src_ip is mapped to x-ibm-finding.src_ip_ref as well as to x-oca-asset.ip_refs.

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: its just redundant to split up the url and reconstruct it for the api call. you can just make api call with the next url here without all this extra steps- `self.client.call_api()`

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: I wonder why you are not using next_url for pagination. the calculation generates all the pagination parameters that are already contains in next_url. The calculation is needed for the first call but for subsequent calls until the result limit you can use next_url. let me know if i am missing something!

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
mdazam1942: But ideally that's part of network-traffic object. I also see `src_host` field in the sample results which contains an IP but don't see it in the mapping though. If its one source to many dest that should also be fine.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: In next_url, we are not doing any calculations, only splitting the next_url because it contains the host name.
When making an API call, we should only pass the API endpoint with parameters. host name value set at the init function.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: Example for the next URL:
'https://host.com/api/v2.4/search/detections/?page_size=100&query_string=query&page=2'

next url contains
host name= host.com 
endpoint= /api/v2.4/search/detections/ 
pagination_parameters= page_size=100&query_string=query&page=2


If we directly use the "self.client.call_api()" function, even then we should split the host and pass the remaining parameters.
Making an api call using self.client.call_api(), we should only pass the endpoint and parameters, not the host name.
While initializing self.client, we are setting the host name.
self.client = RestApiClientAsync(connection.get('host'), port=None, headers=self.headers)

It is similar to the way we are handling in next_url currently.

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
thangaraj-ramesh: In case of multiple network-traffic objects from a list of corresponding data source dict, 
when we tried to map, it(ipv4-addr object created for src_ip) was added as reference to first network-object only due to stix-shifter framework constraint.
src_host is mapped to x-oca-asset. src_ip is referenced to x-oca-asset.ip_refs from outside src_host , being same.

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: you can keep the current implementation to handle pagination api call.

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
mdazam1942: as discussed its fine to reference src_ip to the first network traffic object.

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: as discussed `metadata` parameter needs to be used in order to save the pagination url and results count.

thangaraj-ramesh: Pagination handled using metadata

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
thangaraj-ramesh: Updated mapping to refer src_ip in network traffic object.

mdazam1942 (approved):

PR #1541, opened 2023-07-26
QRADAR v4.X - Remove SNI

KaneBrennan132:
Removing SNI configuration from QRadar 4.X module. Users have been pointing out that the SNI field has been removed from the QRadar assets connector configuration, and not the configuration for the connection profile for UDI. This is due to SNI only being removed in versions 5.X and not versions 4.X (What they see and use). This is causing confusion for users and bugs are being raised around this.


delliott90:
Note needed for this branch.

PR #1559, opened 2023-08-15
Fix ECS range queries with x-oca-event:start/end

Technici4n:
For some reason, these fields were converted from a timestamp to a number representing milliseconds since epoch, messing up the conversion to Lucene range queries when used with comparison operators.

(For example the test would produce a query with `event.start:<=1702339200000` which is not valid Lucene syntax).


pcoccoli:
Are you sure `event.start:<=1702339200000` is not valid?
https://www.elastic.co/guide/en/elasticsearch/reference/8.5/query-dsl-query-string-query.html#_ranges

pcoccoli (approved):
Looks correct to me

mdazam1942 (approved):

PR #1560, opened 2023-08-15
Rework ECS translation to always produce a single query

Technici4n:
Splitting the STIX query into multiple ECS queries when qualifiers are encountered does not make sense with combining operators such as `AND` and `OR`. Instead, we now process START/STOP qualifiers on-the-fly.

Also fix `AND` between observation expressions being converted to `OR`.

The output is still an array of queries, but always of length 1.


stix_shifter_modules/elastic_ecs/stix_translation/query_constructor.py
```diff
@@ -156,62 +154,27 @@ def _parse_expression(self, expression, qualifier=None) -> str:
...
+            return self._parse_expression(expression.comparison_expression)
+        elif isinstance(expression, StartStopQualifier):
+            return "{} AND {}".format(self._parse_expression(expression.observation_expression), _format_start_stop_qualifier(expression))
         elif isinstance(expression, Pattern):
             return "{expr}".format(expr=self._parse_expression(expression.expression))
         else:
             raise RuntimeError("Unknown Recursion Case for expression={}, type(expression)={}".format(
                 expression, type(expression)))
 
-    def parse_expression(self, pattern: Pattern):
```
Technici4n, line 212: Is this considered part of the public API? And if so should it be added back with a deprecation warning?

Technici4n: Added back without a warning as it matches the structure of the other modules.

pcoccoli:
I'm not sure this change is correct.  A STIX observation expression (all the stuff in a set of square brackets) means that all comparison expressions within it need to match the same "observation" (event or record in the case of stix-shifter; "document" in elasticsearch terms).  When you `AND` two observation expressions together in a STIX pattern, you're basically specifying 2 *different* queries at the same time.  This makes sense when describing a particular threat; you're saying if both expressions are matched then that threat is present.  In terms of querying or information retreival, it's a bit trickier.  Should the system return all matches, regardless of which expression it matched, or (when using `AND`) should the system *only* return the matches f there's at least 1 match to both expressions.  stix-shifter chose to do the former, and this is why that use of `AND` gets converted to `OR`.

As an example, look at the unit test `test_query_from_multiple_observation_expressions_joined_by_AND`.  The STIX pattern is `[url:value = 'www.example.com'] AND [mac-addr:value = '00-00-5E-00-53-00']`.  This is 2 expressions `AND`ed together, so what it's looking for is at least one event with that URL and *a different* event with that MAC addr.  Note that this is *not* the same as the STIX pattern `[url:value = 'www.example.com' AND mac-addr:value = '00-00-5E-00-53-00']` (notice this one is a single observation expression).  This pattern looks for a *single* event containing *both* of those values (technically this is invalid in STIX, but stix-shifter allows it).

Technici4n:
Thanks for the explanation @pcoccoli, that is not intuitive but it makes sense! Moving everything to the inside of the `[]` fixed my use case. :)

Issue #1569, opened 2023-08-29
case insensitive support for regex in elastic_ecs


subbyte:
**Is your feature request related to a problem? Please describe.**
Currently, the elastic_ecs connector does not support case insensitive flag like `/abc/i` or `/(?i)abc/`, and hunters need to write `command_line MATCHES '([Vv][Ii][Rr][Uu][Ss])'`.

**Describe the solution you'd like**
From syntax, we expect users to write `command_line MATCHES '(?i)virus'`. After checking more into Elasticsearch docs, we found the most feasible solution to do it in stix-shifter is to unfold `(?i)` in translation and keep the transmission as current implementation.

PR #1573, opened 2023-09-05
resolve case insensitive regex #1569

subbyte:


stix_shifter_modules/elastic_ecs/stix_translation/query_constructor.py
```diff
@@ -304,6 +305,77 @@ def _format_translated_queries(query_array):
     return formatted_queries
 
 
+def _unfold_case_insensitive_regex(regex_pattern):
+    # this function should be executed after "\s" unfolding
+
+    if "(?i)" in regex_pattern:
+
+        escaped_left_bracket_symbol = "辶"
```
delliott90, line 313: Are these the characters you meant to include? They don't look alphanumeric.

subbyte: Yes, this is a hack to distinguish square bracket and escaped square bracket in a regex. I intentionally use the non-ascii unicode symbols, which should not exist in any dataset (partial Chinese char, even not a valid Chinese char).

subbyte: More background: chars inside square bracket of a regex should be handled differently than chars outside square brackets. However, recognizing square bracket is tricky since it may be escaped, and the escape symbol (backslash) may also be escaped so `\[` may not mean escaped square bracket if the char before these substring is `\` like `hello\\[hi`. The unicode with split/join is the hack to deal with these.

delliott90 (approved):

PR #1581, opened 2023-09-26
Vectra config changes

thangaraj-ramesh:
Vectra config changes
   Added port
   Removed result limit as 10000, since vectra supports higher value.


stix_shifter_modules/vectra/configuration/config.json
```diff
@@ -1,22 +1,21 @@
 {
     "connection": {
         "type": {
-            "displayName": "Vectra",
-            "group": "vectra"
```
delliott90, line 5: You should keep the group here.

stix_shifter_modules/vectra/stix_transmission/connector.py
```diff
@@ -223,6 +223,16 @@ def get_results_data(self, response_dict):
...
 
+            # adding new field for confidence field mapping
+            if 'certainty' in record:
+                record['certainty_score'] = record['certainty']
+
+            # converting 0 to 1 for severity and confidence
+            if record.get('certainty_score') == 0:
+                record['certainty_score'] = 1
+            if record.get('threat') == 0:
+                record['threat'] = 1
```
mdazam1942, line 234: this change doesn't make sense. why are you altering the value?

stix_shifter_modules/vectra/stix_transmission/connector.py
```diff
@@ -223,6 +223,16 @@ def get_results_data(self, response_dict):
 
             detection_type = record.get('detection_type', '')
 
+            # adding new field for confidence field mapping
+            if 'certainty' in record:
+                record['certainty_score'] = record['certainty']
```
mdazam1942, line 228: can you please explain this change?

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942, line 234: put some detail comments about this change. for example: the reason for altering the value

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
```diff
@@ -52,18 +52,6 @@
...
-      "object": "detection"
-    },
-    "num_attempts": {
-      "key": "x-ibm-finding.event_count",
-      "object": "detection"
-    },
-    "num_successes": {
-      "key": "x-ibm-finding.x_num_successes",
-      "object": "detection"
-    },
```
mdazam1942, line 66: is a transformer not working for these field values?

thangaraj-ramesh: Transformer isn't applicable as there is no conversion required. 
This attribute is specific to a particular detection type from data source only.
If query results in number of detections where at least one detection consists of this attribute then it gets added to those detection types where it is not present, with value 'NaN' in STIX output.
And this happens from CP4S side. In CLI,  this attribute is present in those particular stix records only.

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: Updated with comments

stix_shifter_modules/vectra/configuration/config.json
thangaraj-ramesh: Added group in config

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942, line 228: do you still need this `certainty_score` field?

thangaraj-ramesh: Yes, certainty_score field is required. 
Previously, the certainty field was mapped to both x-ibm-finding:confidence and x-ibm-ttp-tagging:confidence fields.
As per the schema definition x-ibm-ttp-tagging:confidence range is 0-1 and x-ibm-finding:confidence range is 1-100.
From the vectra datasource, we are getting confidence value in the range 0-100.
x-ibm-ttp-tagging:confidence requires a transformer to convert confidence value in the range 0-100 to 0-1.
Since the same field was mapped to two different stix fields,  when transformer was applied to one field it was applied to both the fields when tested from CP4S.
Hence to avoid this, different field certainty_score is created for mapping x-ibm-finding:confidence.

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
```diff
@@ -103,16 +104,18 @@
     }
   ],
   "certainty": [
-    {
-      "key": "x-ibm-finding.confidence",
-      "object": "detection"
-    },
```
mdazam1942, line 109: this mapping looks valid. not sure why you need to remove this and put it in `certainty_score` field?

stix_shifter_modules/vectra/stix_transmission/connector.py
mdazam1942: i dont think you need this when you revert the mapping change. see https://github.com/opencybersecurityalliance/stix-shifter/pull/1581/files#r1348755907

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
thangaraj-ramesh: Previously, the certainty field was mapped to both x-ibm-finding:confidence and x-ibm-ttp-tagging:confidence fields as given below: 
"certainty": [
    {
      "key": "x-ibm-finding.confidence",
      "object": "detection"
    },
    {
      "key": "x-ibm-ttp-tagging.confidence",
      "object": "ttp_finding",
      "transformer": "ConvertToReal"
    }
  ],

As per the schema definition x-ibm-ttp-tagging:confidence value is in range 0-1 and x-ibm-finding:confidence value is in the range 1-100.
From the Vectra Datasource, certainty value is coming in the range 0-100. Hence x-ibm-ttp-tagging:confidence requires a transformer to convert certainty value from the range [0-100] to [0-1].
Since the same field certainty was mapped to two different STIX fields x-ibm-ttp-tagging:confidence and x-ibm-finding:confidence, when the transformer “ConvertToReal“ was applied to x-ibm-ttp-tagging:confidence it was applied to both the fields x-ibm-ttp-tagging:confidence and x-ibm-finding:confidence when tested from CP4S.
Hence to work around this issue from CP4S, a copy of certainty field was made as certainty_score. Field certainty_score was mapped to x-ibm-finding:confidence and the issue was resolved .

stix_shifter_modules/vectra/stix_transmission/connector.py
thangaraj-ramesh: As per your suggestion removed the code

stix_shifter_modules/vectra/stix_translation/json/to_stix_map.json
thangaraj-ramesh: As per your suggestion removed the code

mdazam1942 (approved):

PR #1637, opened 2024-01-15
Add query batchsize(length) in common config.json

mdazam1942:
We set the default value to 2000 and make it optional. Connector may not need to use this parameter unless it is necessary to enforce a lower query length. The caller application or script can use the value to make the query when it is necessary. The max value is arbitrary and can be changed to a different value is required.


DerekRushton (approved):
Azam told me too.

Issue #1633, opened 2024-01-08
SSL error in stix-bundle connector (v7)


subbyte:
**Describe the bug**
Encountered an error when integrating stix-shifter v7 into Kestrel.

Looks like the stix-bundle transmission module hit an error with SSL even with `ping` command.

Interestingly, this error happens with Python 3.10 and 3.11, but not with 3.8 and 3.9.
- https://github.com/opencybersecurityalliance/kestrel-lang/actions/runs/7450276292/job/20269128127?pr=457
- https://github.com/opencybersecurityalliance/kestrel-lang/actions/runs/7450276292/job/20269129098?pr=457

**To Reproduce**
```
% stix-shifter transmit stix_bundle '{"host": "https://github.com/opencybersecurityalliance/data-bucket-kestrel/blob/main/stix-bundles/lab101.json?raw=true"}' '{"auth": {"username": "", "password": ""}}' ping
 2024-01-08 11:48:53,532 stix_shifter_utils.utils.error_response ERROR    stix_bundle connector error occurred: client_connector_error: (Cannot connect to host github.com:443 ssl:<ssl.SSLContext object at 0x104becb90> [Cannot create a client socket with a PROTOCOL_TLS_SERVER context (_ssl.c:809)])
{
    "success": false,
    "connector": "stix_bundle",
    "error": "stix_bundle connector error => client_connector_error: (Cannot connect to host github.com:443 ssl:<ssl.SSLContext object at 0x104becb90> [Cannot create a client socket with a PROTOCOL_TLS_SERVER context (_ssl.c:809)])",
    "code": "service_unavailable"
}
```

**Expected behavior**
No error.

**Desktop (please complete the following information):**
 - Python 3.11.6
 - stix-shifter 7.0.1

subbyte:
When running `pytest` for all tests under Python 3.11.5, I get two failures (same one as above):
```
stix_shifter_modules/stix_bundle/test/test_stix_bundle_transmission.py F..F.

============================================================================================== FAILURES ==============================================================================================
_________________________________________________________________________________ TestSTIXBundleConnector.test_ping __________________________________________________________________________________

self = <test_stix_bundle_transmission.TestSTIXBundleConnector testMethod=test_ping>

    def test_ping(self):
        entry_point = EntryPoint(self.connection, self.configuration)
        ping_result = run_in_thread(entry_point.ping_connection)
>       assert ping_result["success"] is True
E       assert False is True

[...]

PR #1664, opened 2024-03-14
Changes to support Python 3.10, 3.11 and 3.12

DerekRushton:


subbyte:
Nice! This is likely to resolve #1633

PR #1668, opened 2024-03-19
sumologic: use milliseconds since epoch for timestamps

### LINKED ISSUE #1659: Wrong Timestamp format in Sumologic Module
**Describe the bug**
Docs: https://help.sumologic.com/docs/api/search-job/#create-a-search-job
The format used in the sumologic module is `YmdTHMS`. The right format (see docs) is `Y-m-dTH:M:S`

Also milliseconds not supported (.000Z)

When trying to execute the query with the wrong time format, I get this error: `"message" : "The 'from' field (null) contains an invalid time."`

pcoccoli:
Fixes #1659


mdazam1942 (changes requested):
i think there's a conflict because of the previous sumo logic pr

mdazam1942 (approved):

PR #1673, opened 2024-04-02
Amazon athena resolve column not found exception

DerekRushton:
1. Fixes an issue where the column list included columns from unselected databases.
2. Added additional logic to handle column not found exception.
3. Resolved an issue where the SQL keyword "group" was causing issue with some mappings in from_stix for oscf.


stix_shifter_modules/aws_athena/stix_translation/json/ocsf_from_stix_map.json
```diff
@@ -637,19 +637,19 @@
         "identity.authorizations.policy.desc"
       ],
       "authorizations.policy_group_desc": [
-        "identity.authorizations.policy.group.desc"
+        "'identity.authorizations.policy.group.desc'"
```
mdazam1942, line 640: why did you add extra single quotes?

DerekRushton: "group" is considered a special keyword in SQL. If you try to query this field without quotes, it will always fail because it's going to interpret it as the keyword. Putting quotes around it, to my knowledge, allows sql to interpret it as intended (as a an object in the table). 

I admittedly may need more testing on this one, I just checked that the query doesn't error out after the change.

stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
```diff
@@ -57,36 +62,72 @@ async def create_query_connection(self, query):
...
+        response_dict = await self.client.makeRequest('athena', 'start_query_execution', **query_args)
+        return_obj['search_id'] = response_dict['QueryExecutionId'] + ":" + query_service_type
+
+        modified_query = dict()
+        modified_query = await PostQueryConnectorErrorHandler.check_status_for_missing_column(self.client, return_obj['search_id'], query)
+        #If the query is successful (or the exception isn't column related) than it's considered a success and exits.
+        #If 10 columns are not found or it fails to replace a column 10 times, than it exits (to prevent endless loops).
+        #If the query is not successful, than it will retry the query with the modified query.
+        if(modified_query == "CONNECTOR_FACTORY_SUCCESS" or self.total_try_count > 10):
+            logger.debug(f"The total attempt count was {self.total_try_count}")
```
mdazam1942, line 118: Add a text that indicates what it is trying to attempt so that it is easily understandable from the log

stix_shifter_modules/aws_athena/stix_transmission/post_query_connector_error_handling.py
```diff
@@ -0,0 +1,112 @@
...
+                    break
+            elif(status_response != None and "status" in status_response and status_response["status"] != "RUNNING"):
+                #If there is no match and the status is not running, than stop trying and exit with the success message.
+                break
+                
+        if(column_to_delete != ""):
+            return PostQueryConnectorErrorHandler._remove_invalid_column_table(column_to_delete, query)
+        else:
+            #May not always be successful, just that no column error is occurring.
+            return "CONNECTOR_FACTORY_SUCCESS"
```
mdazam1942, line 42: i would return a boolean True instead of this odd looking string. you can change the function declaration like this:

`async def check_status_for_missing_column(client, search_id, query) -> None:`

stix_shifter_modules/aws_athena/stix_transmission/query_connector.py
```diff
@@ -57,36 +62,72 @@ async def create_query_connection(self, query):
...
+        query_args = {"QueryString": query, "ResultConfiguration": result_config}
+        response_dict = await self.client.makeRequest('athena', 'start_query_execution', **query_args)
+        return_obj['search_id'] = response_dict['QueryExecutionId'] + ":" + query_service_type
+
+        modified_query = dict()
+        modified_query = await PostQueryConnectorErrorHandler.check_status_for_missing_column(self.client, return_obj['search_id'], query)
+        #If the query is successful (or the exception isn't column related) than it's considered a success and exits.
+        #If 10 columns are not found or it fails to replace a column 10 times, than it exits (to prevent endless loops).
+        #If the query is not successful, than it will retry the query with the modified query.
+        if(modified_query == "CONNECTOR_FACTORY_SUCCESS" or self.total_try_count > 10):
```
mdazam1942, line 117: you can just check the boolean value instead of checking a string if you return a boolean true

mdazam1942 (approved):

PR #1683, opened 2024-04-23
Reaqta various mapping fixes

DerekRushton:


DerekRushton:
Explaining the changes : 

1. Mitre tactics came as [1,2,3,4] where each number represented a tactic. The primary issue is that we require that the tactic_name must be a string and that it's required.

There were two possible solutions, we either map on the technique and change tactic_name to represent all the tactics, or we can create multiple mitre objects for each tactic. I've opted for the later to adhere to the extensions requirements.

Hence I've transformed the array into a dict object where each object has the technique, tactic_name (using the map from reaqta UI) and the tactic_number as an extension. Then the TTP references are grouped up for each tactic.

DerekRushton:
Explaining the changes :

3. There existed an asset object that didn't quite make sense. It was always matching on the hostname_id which didn't really mean it was an asset. It was mapping some IP's to it, but there were no garuntee that IP's would be included. It was possible for their to be a host_id with no other fields.

I've changed this to no longer have that asset and instead map those fields to a reaqta extension object. There were other asset objects that made more sense.

stix_shifter_modules/reaqta/test/stix_translation/test_reaqta_json_to_stix.py
```diff
@@ -506,4 +479,51 @@ def test_windows_process_event(self):
...
+        data = {'eventId': '1119732427405660161', 'endpointId': '1114241789272784896', 'payload': {'localId': '1119732154209667073', 'process': {'id': '1114241789272784896:4704:1713564948917', 'parentId': '1114241789272784896:4060:1713564802263', 'endpointId': '1114241789272784896', 'program': {'path': 'c:\\users\\test\\desktop\\maze_rw.exe', 'filename': 'maze_rw.exe', 'md5': '01010101010101010101010101010101', 'sha1': '0101010101010101010101010101010101010101', 'sha256': '010101010101010101010101010101010101010101010101010101010101010', 'size': 495616, 'arch': 'x32', 'fsName': 'maze_rw.exe'}, 'user': 'USER-Desktop\\Desktop', 'pid': 4704, 'startTime': '2024-04-19T22:15:48.917Z', 'ppid': 4060, 'pstartTime': '2024-04-19T22:13:22.263Z', 'userSID': 'S-1-5-21-0101010101-0101010101-0101010101-1001', 'privilegeLevel': 'HIGH', 'noGui': False, 'logonId': '0x14c9e0'}, 'incidents': [], 'triggeredIncidents': ['1119732154268387330'], 'data': {'matched': [{'policyId': '1072556371783712772', 'versionId': '1072556371783716869', 'matcherId': '1072556371783720966'}]}, 'eventType': 28}, 'happenedAt': '2024-04-19T22:15:49.181Z', 'receivedAt': '2024-04-19T22:16:54.316Z'}
```
mdazam1942, line 485: i would put the sample in stix_shifter_modules/reaqta/test/stix_translation/json for easier maintainability

stix_shifter_modules/reaqta/stix_transmission/connector.py
```diff
@@ -121,6 +142,21 @@ def modify_result(cls, data):
...
                 result['payload'] = cls.update_net_traffic_flow(payload, network_protocol)
+            
+            if result.get('payload') and result['payload'].get('data') and result['payload']['data'].get('tactics'):
+                tactic_list = result['payload']['data']['tactics']
+                technique = result['payload']['data']['technique']
+                result['payload']['data']['mod_tactics'] = {}
+                dict_list = []
+                for tactic in tactic_list:
+                    
+                    if (f"{tactic}" in TACTIC_NAME_MAPPING):
```
mdazam1942, line 153: `TACTIC_NAME_MAPPING.get(tactic)` more readable?

maybe you wanna change the dictionary name to simple lowercase

mdazam1942: you can save it in `stix_shifter_modules/reaqta/stix_translation/json` and read it from there. but that would be a overkill. this is already hard coded in the code. so no one will be able to change it anyway. you can probably add a debug log and add something in the comment

stix_shifter_modules/reaqta/test/stix_translation/test_reaqta_json_to_stix.py
DerekRushton: Made the change.

stix_shifter_modules/reaqta/stix_transmission/connector.py
DerekRushton: I'll just modify it to use the lower case. There is already a comment describing it's purpose.

mdazam1942: `tactic_name_mapping.get(tactic)` ?

stix_shifter_modules/reaqta/stix_transmission/connector.py
```diff
@@ -150,8 +150,8 @@ def modify_result(cls, data):
                 dict_list = []
                 for tactic in tactic_list:
                     
-                    if (f"{tactic}" in TACTIC_NAME_MAPPING):
-                        dict_list.append({"tactic_number": tactic, "tactic_name": TACTIC_NAME_MAPPING.get(f"{tactic}"), "technique":technique})
+                    if (f"{tactic}" in tactic_name_mapping):
+                        dict_list.append({"tactic_number": tactic, "tactic_name": tactic_name_mapping.get(f"{tactic}"), "technique":technique})
```
mdazam1942, line 154: is there a reason you need to f'string the tactic value?

stix_shifter_modules/reaqta/stix_transmission/connector.py
DerekRushton: Yes, that's what I've left it as.

stix_shifter_modules/reaqta/stix_transmission/connector.py
DerekRushton: The mapping was in string, but the values were int. I've changed the mapping to match what the API responds with (int) so I won't need the f string.

PR #1696, opened 2024-06-03
Create e2eStixBundle01.json

dlavigneNBguy:
add new file for e2e testing , needed to have ipv6 and user observables

### DIFF
--- a/data/cybox/e2eStixBundle01.json
+++ b/data/cybox/e2eStixBundle01.json
@@ -43,7 +43,7 @@
           },
           "5": {
             "type": "user-account",
-            "value": "isc-demo"
+            "value": "e2e user"
           }
         }
       }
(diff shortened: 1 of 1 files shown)


benjamin-craig (approved):

data/cybox/e2eStixBundle01.json
```diff
@@ -0,0 +1,51 @@
...
+            "type": "ipv6-addr",
+            "value": "0000:0000:0000:0000:0000:0000:0000:0001"
+          },
+          "4": {
+            "type": "url",
+            "value": "http://kbfvzoboss.bid/alien/fre.php"
+          },
+          "5": {
+            "type": "user-account",
+            "value": "isc-demo"
```
DerekRushton: This probably should be a fake value. Can you ensure that the rest of the values are fake.

dlavigneNBguy: yes, all values are fake, technically isc-demo is fake, it is a real account but there is no "real" person associated.  Ill change to something else regardless though.

DerekRushton (approved):

PR #1727, opened 2024-08-26
Added the ability to enable trusted_env.

DerekRushton:


stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, host, port=None, headers={}, url_modifier_function=None, cert
         unique_file_handle = uuid.uuid4()
         self.server_cert_name = "/tmp/{0}-server_cert.pem".format(unique_file_handle)
         server_ip = host
+        self.trust_env_enabled = os.environ.get("STIX_SHIFTER_ENABLE_TRUST_ENV", False)
```
benjamin-craig: Should we maybe call this something with PROXY?

stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, host, port=None, headers={}, url_modifier_function=None, cert
...
+        self.logger.error(f"Proxy Environment - Trusted_Env Enabled : {self.trust_env_enabled}")
```
benjamin-craig: Should this be info vs error?

benjamin-craig (approved):

stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
DerekRushton: I've updated it to debug.

stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
```diff
@@ -57,6 +57,17 @@ def __init__(self, host, port=None, headers={}, url_modifier_function=None, cert
         unique_file_handle = uuid.uuid4()
         self.server_cert_name = "/tmp/{0}-server_cert.pem".format(unique_file_handle)
         server_ip = host
+        
+        #To enable proxy, set the environment variable "STIX_SHIFTER_ENABLE_TRUST_ENV" to true. This option will allow the connection
+        #to use the system environments proxy settings. This can be done by setting the "https_proxy" environment variable to 
+        #"http(s)://[username]:[password]@[hostname]/[ipaddress]:[port]". Alternative proxy schema's may or may not work.
+        self.trust_env_enabled = os.environ.get("STIX_SHIFTER_ENABLE_TRUST_ENV", False).lower()
```
benjamin-craig: Do you think STIX_SHIFTER_ENABLE_TRUST_PROXY_ENV would be more descriptive or just leave it as it is ?

benjamin-craig (approved):

stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
benjamin-craig: I think we need to remove the .lower there.

stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py
DerekRushton: I wanted it to match the actual setting we are configuring. It does only impact proxy and it is only being used for that, but I prefer it to be clear of what it is changing.

PR #1749, opened 2024-12-23
Securonix :: STIX Shifter

raghul-simbian:
This connector is still in dev. Testing in progress


DerekRushton:
Can you please include a write-up about what this connector is supposed to do. I'd ideally be looking for 
1. A summary of what Securonix is and how it works.
2. What kind of data do you expect to get back from this connector.
3. Sample queries and Sample output. 
4. A high level description of how it pulls the data in. What types of authentication are used, how many queries, etc.
5. We do require at least 80% code coverage and unit testing.
6. Any other information would be appreciated.

I may not get back to you, but I'll try. However whoever will own this repo going forward would likely need this information as well. Opening this as an issue would be a good place to put that conversation so we can figure out if this is a good fit for the repo.

PR #1769, opened 2025-11-18
Migrate Away from Invoking Setup.py (i.e., Build System Refactor for PEP 517)

morgan-nolan-ibm:
### Summary

This PR refactors the STIX-Shifter build process by replacing the deprecated `setup.py` workflow with a **[PEP 517](https://peps.python.org/pep-0517/)/[518](https://peps.python.org/pep-0517/)-compliant build system**. The previous use of `setup.py` was no longer supported, causing builds to fail. This update restores build functionality while improving readability and maintainability.

---

### Changes

- Introduced a new module, `build_tools`, to handle building and packaging. It includes:
  - **`run_build.py`**: Main driver for the build process. Supports optional wheel installation (`install`) and additional arguments (e.g., `--no-isolation`) for customizing the underlying `python -m build` call.
  - **`build.py`**: Core build logic that preserves the original behavior of `setup.py`.
  - **`pre_build.py`**: Installs build requirements and prepares the environment.
  - **`generate_pyproject.py`**: Dynamically generates `pyproject.toml` files using a template.
  - **`logging_setup.py`**: Provides logging configuration for the build_tools module.

- Replacement commands for `setup.py`:
  - `python -m build_tools.run_build`: runs pre-build steps and builds wheels.
  - `python -m build_tools.run_build install`: builds and installs wheels.
  - `python -m build_tools.pre_build`: installs requirements only.
  - `python -m build_tools.run_build [install] [additional_args]`: allows optional installation and passing extra arguments (e.g., `--no-isolation`) to the underlying build command.

#### Additional Additions
- Build code is refactored to be more maintainable and debuggable.
  - Code is now seperated into individual functions according to responsibility and comments have been added describing their purpose.
  - Wheels are now pre-packaged in a temp folder that can be preserved by toggling function variable.
  - Robust logging has now been added, will respect log-levels.
    - Recommended to use DEBUG logs on feature builds, and probably INFO where there is confidence that nothing is breaking.

---

### Motivation

- `setup.py` was deprecated and direct invocation is no longer supported.
- Builds stopped working due to this deprecation.
- Implementing [PEP 517](https://peps.python.org/pep-0517/)/[518](https://peps.python.org/pep-0517/)-compliant builds ensures compatibility with modern Python packaging tools and future-proofs the project.

---

### Testing

- Verified that wheels can be built and installed using the new commands.
- Ensured that the generated `pyproject.toml` files are valid and usable.
- Ensured wheels maintained expected structure while development was ongoing.
- CI/CD pipeline successfully builds with the new system.
- Deployed to cluster and observed no abnormality.

---

### Notes

- All references to `setup.py` have been updated to use the new `build_tools` commands.  
- Some calling scripts still include redundant installations of `setuptools` or other build dependencies. These can be cleaned up in the future, as the new build system installs all required packages (`setuptools`, `wheel`, and others specified in `pyproject.toml`) in an isolated environment by default. Furthermore, setuptools is now included in our dev requirements, as we require import outside of running `python -m build ...` for `find_packages` import in our main build script.  
- The new commands (`python -m build_tools.run_build` and `python -m build_tools.run_build install`) handle building and installation in a consistent, isolated way, so explicit dependency installs in scripts are no longer necessary for the build itself.  
- A **setuptools warning** is currently introduced due to ambiguity when adding configuration files (e.g., `.json`) manually to the project. Setuptools prefers these paths to be specified in `pyproject.toml`. This is a known item for future work and is unlikely to cause issues with current builds.  
- All other deprecations have been addressed. Two additional `setup.py`-related deprecations originally scheduled for February and March have now been resolved with this refactor.  

---

## Compatibility Notes

- wheel name produced now follow `NAME_VERSION-py3-none.whl` naming scheme rather than `NAME_VERSION-py2-py3-none-any.whl`. Impact is that wheels no longer contain metadata alluding to support for Python `v2.x`.
- Calling scripts referencing `setup.py` will break now that it has been deleted and replaced with `python -m build_tools.pre_build`.
- `generate_requirements.py` has been kept for now in root folder, but `python -m build_tools.pre_build` or setting `INSTALL_REQUIREMENTS_ONLY` and running `python -m build_tools.run_build` is suggested command to be used going forward.

---

### References

- [Setup.py Deprecation](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).  
- [Some Q&As on this page are informing in regards to this work](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/)


morgan-nolan-ibm:
Here's a generated report comparing the wheels built locally to the latest ones pulled from PyPI. All changes look expected.

**Why the validation report is longer than one would expect:**  
The serialization order of the configuration files appears to be random. This is likely due to `sort_key=False` and the use of `os.walk` during data discovery. This doesn’t affect functionality, the JSON data is still read and interpreted correctly, and data integrity is fully preserved. The difference is only in ordering. My build code uses the existing construction functions, so this behaviour should be consistent with how the original implementation generated these configuration files.

[wheel_validation_report.txt](https://github.com/user-attachments/files/23878894/wheel_validation_report.txt)

Kieran7741 (approved):

Peter-Sheehan (approved):
I have limited knowledge of this codebase but from a high level everything looks good.

Issue #1493, opened 2023-05-19
v5 RestApiClientAsync much slower than v4 RestApiClient and cause connect_timeout


subbyte:
**Describe the bug**
The `aiohttp` or `aiohttp-retry` library in stix-shifter v5 looks very slow and cause `connect_timeout`, which is not an issue in stix-shifter v4.

**To Reproduce**
Run the STIX query `[ipv4-addr:value = '127.0.0.1']` or the following equivalent Kestrel huntflow:
```
ips = GET ipv4-addr FROM stixshifter://abc
      WHERE value = '127.0.0.1'
      START 2022-10-01T00:00:00Z STOP 2023-05-01T00:00:00Z
```
We have more than 55k data in an Elasticsearch store for this query to retrieve.

[note] STIX query triggered by the stix-shifter CLI
- does not have pagination
- does not retrieve all results
- so it does not reveal the issue described below in a real app

The Kestrel huntflow with the following setting
```
            options:
                result_limit: 500000
                retrieval_batch_size: 10000
                timeout: 60
```
will be able to retrieve all data in 5 batches. The rest api `call()` in [stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py](https://github.com/opencybersecurityalliance/stix-shifter/blob/develop/stix_shifter_utils/stix_transmission/utils/RestApiClientAsync.py#L122) will be invoked 6 times---1 for page length probing, 5 for `transmit.results()`.

A few observations:

1. When print out each time the `call()` function is invoked and finished, I see significant more time for it to run in v5 compared with v4. This results in twice amount of time to execute the Kestrel huntflow with stix-shifter v5 than v4.
2. When the 6 `call()` API are executed in a sequence, each one takes more time than the previous `call()`. The time consumed plateaus at some point, e.g., after 4 `call()`.
3. At some point during the 6 `call()` API execution, we could see an error message
```
STIX-shifter transmission.results() failed with message: elastic_ecs connector error => server timeout_error (2 sec)
```
4. The time to execute a `call()` function changes when I change `retrieval_batch_size`---it tells Elasticsearch how many documents/records to return---This cause the error to happen or not to happen (since it is thresholding on time). Smaller `retrieval_batch_size` takes less time to execute the `call()` function and have less chance to emit the `connect_timeout` error. I thought the `connect_timeout` error is not for entire `timeout`, but just for connection establishment, and this parameter in the HTTP request should not affect anything there. So I am confused.

[...]

PR #1772, opened 2026-01-14
Increase default connection timeout to 10 seconds

ibm-app-id-o-fletch:
Increase the `CONNECT_TIMEOUT_DEFAULT` from `2` to `10` as we have had a number of clients who's requests are timing out.  I understand this value can be overridden with the `STIXSHIFTER_CONNECT_TIMEOUT` environment variable however I think `2` is probably too short by default and should be increased.


morgan-nolan-ibm (approved):

morgan-nolan-ibm:
Am linking this issue to PR as a note so it does not get lost, but this Issue Comment describes a better solution to resolving these timeout issues:
https://github.com/opencybersecurityalliance/stix-shifter/issues/1493#issuecomment-1646027822

I don't see an issue with approving this PR though, but should be seen as a mitigation rather than a fix.

